PR #73, opened 2020-06-11
Update azure-pipelines-1.yml for Azure Pipelines

nithyaganeshng:


symorton:
Can you delete replace the azure-pipelines.yml file with this new yaml file (azure-pipelines-1.yml), so that we are not running two different pipelines?

nithyaganeshng:
> Can you delete replace the azure-pipelines.yml file with this new yaml file (azure-pipelines-1.yml), so that we are not running two different pipelines?

Sure, will do.

nithyaganeshng:
> Can you delete replace the azure-pipelines.yml file with this new yaml file (azure-pipelines-1.yml), so that we are not running two different pipelines?

Renamed the file and added lint task.

PR #77, opened 2020-06-15
Adjust metaphor (remove Card)

AndiMiko:
Removed and renamed further references to Cards/PICs. Mainly we have a `VerifiableCredentialHolder` class now, abbreviated VCH.

We still have a references that I can't change because they depend on the service model. `picId` in `VerifiableCredential`. @symorton could you take care of renaming this in the long run if there are any changes going on?

**Type of change:**
- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [X] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:
- [X] Medium – Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)


README.md
```diff
@@ -124,14 +124,15 @@ interface PresentationRequestValidator {
 > note: [OIDC Self-Issued Protocol](https://openid.net/specs/openid-connect-core-1_0.html#SelfIssued) is the only protocol we support as of now in `OidcRequestValidator` class.
 
 
-### Portable Identity Card Data Model
-Portable Identity Cards are simply a container for verifiable credentials and are comprised of:
+### VerifiableCredential Data Model
+VerifiableCredentialHolder is container for verifiable credentials and are comprised of:
```
symorton, line 128: nit: VerifiableCredentialHolder is a container

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -32,11 +32,11 @@ import javax.inject.Singleton
...
 /**
  * This class manages all functionality for managing, getting/creating, presenting, and storing Verifiable Credentials.
- * We only support OpenId Connect Protocol in order to get and present Portable Identity Cards.
+ * We only support OpenId Connect Protocol in order to get and present Verifiable Credentials.
  */
 @Singleton
-class CardManager @Inject constructor(
-    private val picRepository: CardRepository,
+class VerifiableCredentialManager @Inject constructor(
+    private val vchRepository: VerifiableCredentialHolderRepository,
```
symorton, line 39: Should it be VerfiableCredentialHolderRepository or just VerifiableCredentialRepository? For example, the network calls just `GET` and `POST` issued verifiable credentials.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -91,13 +88,10 @@ class CardManager @Inject constructor(
      * Get Issuance Request from a contract.
      *
      * @param contractUrl url that the contract is fetched from
-     *
-     * @return Result.Success: IssuanceRequest object containing all metadata about what is needed to fulfill request including display information.
-     *         Result.Failure: Exception explaining what went wrong.
```
symorton, line 96: why are you deleting all return statements? Just curious?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/contracts/VcContract.kt
```diff
@@ -9,12 +9,12 @@ import com.microsoft.did.sdk.credential.service.models.contracts.display.Display
...
 /**
- * A logical grouping of documents created by an issuer to enable the creation of a Portable Identity Card.
- * In the Portable Identity Card Service, there are four files that make up a contract:
+ * A logical grouping of documents created by an issuer to enable the creation of a Verifiable Credential and meta-data.
+ * In the Verifiable Credential Service, there are four files that make up a contract:
  * schema, display, and input.
  */
 @Serializable
-data class PicContract(
+data class VcContract(
```
symorton, line 17: Should we just call this Contract? Since it is in the context of credential.service.models?

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/SdkDatabase.kt
```diff
@@ -24,11 +24,11 @@ import com.microsoft.did.sdk.datasource.db.dao.VerifiableCredentialDao
  * More info:
  * https://developer.android.com/topic/libraries/architecture/room
  */
-@Database(entities = [VerifiableCredential::class, PortableIdentityCard::class, Identifier::class, Receipt::class], version = 1)
+@Database(entities = [VerifiableCredential::class, VerifiableCredentialHolder::class, Identifier::class, Receipt::class], version = 1)
```
symorton, line 27: Update version number? Well, there will be a merge conflict with Nithya's PR so will have to update anyways.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
AndiMiko: 1. There are some comments that are outdated and are wrong
2. Comments should add useful information, or not be present (otherwise at best they are repeating what has already been said, or at worst giving wrong information). 

In this particular case you might see that every `Result.Failure` is saying the same thing. It is not saying anything other than what the IDE is already telling you through it's type system (it's a bit of a different story for untyped languages). You already know that a Result can be a Failure and you already know that a Failure contains an Exception. There is no added value in my opinion.

For the Success case it's similar. The typesystem already tells you it's a `IssuanceRequest` and then the comment is a rough description of what an `IssuanceRequest` is. This description should be maintained on the class itself instead of here (if need be).

An example for a "useful" class comment would be something that describes specific states that may be dependent on the specific inputs (made up):
"IssuanceRequest: The request in state A if service available, otherwise the latest cached version Z. Depending on the Service meta-data B, C might be present. For successful completion a set of meta-data B, D is required and as such has to be provided by the client. 

As Uncle Bob says "Comments are lies in disguise" :P

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
AndiMiko: That's a good point I actually stumbled over myself but was too lazy to fix lol. I think this repository class should be split up in multiple respositories (for VCs, VCH and Receipts). 

One of us can do that in an upcoming PR? :D

README.md
AndiMiko: haha, true

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/contracts/VcContract.kt
AndiMiko: I see. Which word do we use when we talk about it? When I was explaining it to Daniel I actually had trouble finding a "declerative" word for it. I said "Contract" but that felt a bit naked, shouldn't it be a "VerifiableCredentialContract" (or VcContract)?

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/SdkDatabase.kt
AndiMiko: yea true let me figure that out

AndiMiko: Actually let's commit like this and Nithyas PR will increase it to v2 including both changes

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/contracts/VcContract.kt
symorton: thats fair. ultimately up to you, but i would go with VerifiableCredentialContract then

AndiMiko: Ok, I changed it to the longer form :)

PR #80, opened 2020-06-16
Bugfix scheme checking for deep link

AndiMiko:
Do you remember how I was ranting about the only thing I dislike in Kotlin? @symorton @nithyaganeshng Here we have a perfect example why it's bad ^^ (implicitly returning the last line of a `{ }` scope)

However, minor fix. I also added a constant for the scheme so that Authenticator can consume that.

**Type of change:**
- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [X] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
@@ -76,11 +76,10 @@ class VerifiableCredentialManager @Inject constructor(
             return Result.Success(serializedToken)
         }
         val requestUri = uri.getQueryParameter("request_uri")
-        return if (requestUri == null) {
-            Result.Failure(PresentationException("Request Uri does not exist."))
-        } else {
-            vchRepository.getRequest(requestUri)
+        if (requestUri != null) {
+            return vchRepository.getRequest(requestUri)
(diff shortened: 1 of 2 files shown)


sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -63,24 +64,22 @@ class VerifiableCredentialManager @Inject constructor(
...
         val url = Uri.parse(uri)
-        if (url.scheme != "openid") {
+        if (url.scheme != Constants.DEEP_LINK_SCHEME) {
             throw PresentationException("Request Protocol not supported.")
         }
         return url
     }
 
     private suspend fun getPresentationRequestToken(uri: Uri): Result<String> {
-        return runResultTry {
```
nithyaganeshng, line 73: was removing runResultTry intentional? Don't we need it to handle exceptions and abort on error?

AndiMiko: abortOnError is not used and this is a private method. The exception is catched in the outer public method. So yes, intentional :)

PR #86, opened 2020-06-26
Niganesh/test java library for secp256k1

nithyaganeshng:
Problem:
Library used for cryptographic operations had two drawbacks.
1. Produced and verified signatures only in normalized lower S-form which is not compatible with other libraries.
2. Required lot of efforts to maintain and test against different architectures

Solution:
Replaced the library used for cryptographic operations to Spongycastle. Now signatures are produced in format consistent with other libraries.

**Validation:**

**Type of change:**
- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:
- [x] High – Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)

**Work Item links**:

**Documentation Links**:


sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
```diff
@@ -54,7 +53,8 @@ object VerifiableCredentialSdk {
         SdkLog.addConsumer(logConsumer)
     }
 
+    //TODO: Remove this telemetry as there is no exceptions based on architectures?
```
symorton, line 56: yea should we remove this method now? @AndiMiko

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
```diff
@@ -86,19 +107,77 @@ class Secp256k1Provider(private val subtleCryptoSha: SubtleCrypto) : Provider()
...
+        val ecSpec = ECNamedCurveTable.getParameterSpec(SECP256K1_CURVE_NAME_EC)
+        val ecPoint = ecSpec.g.multiply(BigInteger(1, privateKey))
+        val pubKeySpec = ECPublicKeySpec(ecPoint, ecSpec)
+        val publicKey = keyFactory.generatePublic(pubKeySpec)
+        return byteArrayOf(Secp256k1Tag.UNCOMPRESSED.byte) + (publicKey as BCECPublicKey).q.xCoord.encoded +
+            publicKey.q.yCoord.encoded
+    }
+
+    //TODO: Confirm if we want to support DER encoding
+    fun encodeToDer(r: ByteArray, s: ByteArray): ByteArray {
```
symorton, line 143: so the default signature encoding is now R||S and we have the option to encode/decode signatures to DER?

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
```diff
@@ -86,19 +107,77 @@ class Secp256k1Provider(private val subtleCryptoSha: SubtleCrypto) : Provider()
...
+        val decoder = ASN1InputStream(bytes)
+        val seq = decoder.readObject() as DLSequence
+        val r = seq.getObjectAt(0) as ASN1Integer
+        val s = seq.getObjectAt(1) as ASN1Integer
+        return Pair(r.positiveValue, s.positiveValue)
     }
 
     override fun onVerify(algorithm: Algorithm, key: CryptoKey, signature: ByteArray, data: ByteArray): Boolean {
+        if (signature.size > 64) {
+            throw SignatureException("Signature is not in R|S format")
```
symorton, line 162: So we will have to decode a DER signature before onVerify is called? What if we do this decoding in this method so we automatically support both?

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
nithyaganeshng: I have the methods in there for conversion to and from DER but haven't really used it. With R|S everything is working now. But have to clarify with all the teams that we do not need DER. If it is something we have to support but not default, will invoke the methods.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
nithyaganeshng: if we support DER, instead of throwing an exception, have to call decodeDER in here.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -9,30 +9,35 @@ import android.util.Base64
...
 import java.nio.ByteOrder
+import java.security.KeyFactory
 import javax.inject.Inject
 
 class EllipticCurvePairwiseKey @Inject constructor() {
 
     fun generate(crypto: CryptoOperations, masterKey: ByteArray, algorithm: Algorithm, peerId: String): PrivateKey {
         //TODO: Verify the list of supported curves. Is K-256 supported?
-        val supportedCurves = listOf("K-256", "P-256K")
+        val supportedCurves = listOf("P-256K")
```
beejones, line 40: please use SECP256K1

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -42,14 +47,24 @@ class EllipticCurvePairwiseKey @Inject constructor() {
         if (supportedCurves.indexOf((algorithm as EcKeyGenParams).namedCurve) == -1)
```
beejones, line 47: how about casing?

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
beejones: you can leave it out for the moment. We need to evaluate when we start doing interoperability.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
```diff
@@ -193,40 +264,30 @@ class Secp256k1Provider(private val subtleCryptoSha: SubtleCrypto) : Provider()
...
-        if (key.type == KeyType.Private) {
-            val keyData = (key.handle as Secp256k1Handle).data
-            if (!NativeSecp256k1.secKeyVerify(keyData)) {
-                throw KeyException("Private key invalid")
-            }
+    // Converts the public key returned by library from byte array to x and y co-ordinates to be used in JWK
+    private fun publicToXY(keyData: ByteArray): Pair<String, String> {
+        return when {
+            // Convert uncompressed hex and hybrid hex formats of public key to x and y co-ordinates to be used in JWK format
+            isPublicKeyUncompressedOrHybridHex(keyData) -> publicKeyToXYForUncompressedOrHybridHex(keyData)
```
beejones, line 271: make sure this is big endian

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
```diff
@@ -86,19 +107,77 @@ class Secp256k1Provider(private val subtleCryptoSha: SubtleCrypto) : Provider()
...
+            s = s.sliceArray(1 until s.size)
+        return r + s
+    }
+
+    private fun createCurveParameters(): ECDomainParameters {
+        val ecParams = ECNamedCurveTable.getParameterSpec(SECP256K1_CURVE_NAME_EC)
+        return ECDomainParameters(ecParams.curve, ecParams.g, ecParams.n, ecParams.h)
+    }
+
+    private fun generatePublicKeyFromPrivateKey(privateKey: ByteArray): ByteArray {
```
logangirvin, line 131: Isn't this the same function in ellipticCurvePairwiseKey? Could we refactor it? Given that this implements Provider, it wouldn't normally be called by those wishing to use ec as a provider, but could also be pulled out to a separate ellipticcurveutil if you want to keep these class signatures clean.
I'd be curious what @beejones or @symorton 's thoughts are on this.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
logangirvin: It could be handy if further signature schemes specify DER. I can't imagine formats doing this but you never know.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
logangirvin: I would say only perform this throw based on the algorithm. The signature format is always specified, so we should try to parse accordingly and if it doesn't match just throw. Given how precise cryptography needs to be, recovering from this sort of error could open us up to vulnerabilities (would need analysis of this to be sure though).

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: Both P-256K and SECP256K1 represent secp256k1?

nithyaganeshng: Does K-256 also represent secp256k1? Does it have to be included as well?

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: checked for casing now.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
nithyaganeshng: Converted it to Big Endian.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
nithyaganeshng: On closer look, there were few more functions common and repeated across the two classes. Moved all those to util file.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
nithyaganeshng: Not using them now but leaving them in if needed later.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
nithyaganeshng: Removed it.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
nithyaganeshng: You mean specify format as an additional parameter to algorithm? Currently since we support only R||S and it will always be 64 bytes, I will do it based on the size. When we add DER support as part of interoperability would check if format is DER or not and convert accordingly.

sdk/src/main/java/com/microsoft/did/sdk/util/EllipticCurveUtil.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.microsoft.did.sdk.util.controlflow.KeyFormatException
+import org.spongycastle.jcajce.provider.asymmetric.ec.BCECPublicKey
+import org.spongycastle.jce.ECNamedCurveTable
+import org.spongycastle.jce.spec.ECPublicKeySpec
+import java.math.BigInteger
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.security.KeyFactory
+
+// Converts the public key returned by library from byte array to x and y co-ordinates to be used in JWK
```
AndiMiko, line 17: nit, Use method comment `/**` instead. The IDE will then show this comment in other places where this method is used.

sdk/src/main/java/com/microsoft/did/sdk/util/EllipticCurveUtil.kt
```diff
@@ -0,0 +1,59 @@
...
+    // uncompressed, bytes 1-32, and 33-end are x and y
+    val x = convertToBigEndian(keyData.sliceArray(1..32))
+    val y = convertToBigEndian(keyData.sliceArray(33..64))
+    return Pair(
+        Base64.encodeToString(x.array(), Base64.URL_SAFE or Base64.NO_PADDING or Base64.NO_WRAP),
+        Base64.encodeToString(y.array(), Base64.URL_SAFE or Base64.NO_PADDING or Base64.NO_WRAP)
+    )
+}
+
+fun isPublicKeyUncompressedOrHybridHex(keyData: ByteArray): Boolean {
```
AndiMiko, line 44: nit, put bool operators in front when multiline
```
fun isPublicKeyUncompressedOrHybridHex(keyData: ByteArray): Boolean {
    return keyData.size == 65
        && (keyData[0] == Secp256k1Provider.Secp256k1Tag.UNCOMPRESSED.byte
        || keyData[0] == Secp256k1Provider.Secp256k1Tag.HYBRID_EVEN.byte
        || keyData[0] == Secp256k1Provider.Secp256k1Tag.HYBRID_ODD.byte)
}
```

sdk/build.gradle
```diff
@@ -121,7 +121,7 @@ dependencies {
     implementation "androidx.security:security-crypto:1.0.0-beta01"
 
     api 'com.google.guava:guava:28.1-android' // TODO: this seems to be required as api for org.bitcoin.jar? Can we avoid this?
-    implementation files('libs/org.bitcoin.jar')
+    implementation "com.madgag.spongycastle:prov:1.58.0.0"
```
AndiMiko, line 124: Please add this to cgmanifest in Authenticator for compliance scan unless our SDK pipeline does support scanning within this month :)

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -5,47 +5,46 @@
...
+import com.microsoft.did.sdk.util.publicToXY
 import java.nio.ByteBuffer
-import java.nio.ByteOrder
+import java.util.Locale
 import javax.inject.Inject
 
 class EllipticCurvePairwiseKey @Inject constructor() {
 
     fun generate(crypto: CryptoOperations, masterKey: ByteArray, algorithm: Algorithm, peerId: String): PrivateKey {
         //TODO: Verify the list of supported curves. Is K-256 supported?
```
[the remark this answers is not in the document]
nithyaganeshng: I am going to confirm with Ronny and add all standard names here. Will create a separate PR for it if there is any change in these names.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
```diff
@@ -35,15 +56,14 @@ class Secp256k1Provider(private val subtleCryptoSha: SubtleCrypto) : Provider()
...
     ): CryptoKeyPair {
-        val seed = ByteArray(32)
         val random = SecureRandom()
-        random.nextBytes(seed)
-        NativeSecp256k1.randomize(seed)
+        val seed = random.generateSeed(32)
+        random.setSeed(seed)
 
         val secret = ByteArray(32)
         random.nextBytes(secret)
```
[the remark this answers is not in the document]
nithyaganeshng: It would automatically seed itself but we couldn't do get seed if we need to save it for deterministic key generation. But that is a separate topic of how we are going to use it there, so will remove this now.

sdk/src/main/java/com/microsoft/did/sdk/crypto/plugins/Secp256k1Provider.kt
```diff
@@ -86,19 +106,66 @@ class Secp256k1Provider(private val subtleCryptoSha: SubtleCrypto) : Provider()
...
+        signingSigner.init(true, privateKeyParams)
+
+        val signature = signingSigner.generateSignature(hashedData)
+        var r = signature[0].toByteArray()
+        var s = signature[1].toByteArray()
+        if (r.size > 32)
+            r = r.sliceArray(1 until r.size)
+        if (s.size > 32)
+            s = s.sliceArray(1 until s.size)
+        return r + s
```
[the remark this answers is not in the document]
nithyaganeshng: Done

sdk/src/main/java/com/microsoft/did/sdk/util/EllipticCurveUtil.kt
nithyaganeshng: Done

sdk/build.gradle
nithyaganeshng: Automatic CG didn't pick it up. So I am going to manually register all our dependencies in our pipeline. Will add it in Authenticator's cgmanifest if for some reason that doesn't get done.

sdk/src/main/java/com/microsoft/did/sdk/util/EllipticCurveUtil.kt
nithyaganeshng: Done

PR #93, opened 2020-07-01
Presentation Request Protocol Change and Validator Testing

symorton:
**Problem:**
QR code was too long, so Relying Parties are not using client_id parameter in QR code or Deep Link urls. 

**Solution:**
- Delete param check in `OidcPresentationRequestValidator` class
- Add tests for both `OidcPresentationRequestValidator` and `JwtValidator` classes using mockk
- Remove dependency on Mockito

**Validation:**
- Unit tests pass

**Type of change:**
- [ ] Feature work
- [X] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [X] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:
[Client ID Bug](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1056213)

**Documentation Links**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/credential/service/validators/OidcPresentationRequestValidator.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/credential/service/validators/OidcPresentationRequestValidator.kt
@@ -27,9 +27,7 @@ class OidcPresentationRequestValidator @Inject constructor(
         if (!jwsValidator.verifySignature(token)) {
             throw ValidatorException("Signature is not Valid.")
         }
-        // TODO(check token expiration when implemented in sdk)
-        // checkTokenExpiration(request.content.exp)
-        checkRequestParameters(request.content, request.uri)
+        checkTokenExpiration(request.content.exp)
     }
 
     private fun checkTokenExpiration(expiration: Long) {
@@ -43,13 +41,7 @@ class OidcPresentationRequestValidator @Inject constructor(
         return currentTimeInSeconds + SECONDS_IN_A_MINUTE * expirationCheckTimeOffsetInMinutes
     }
 
-    private fun checkRequestParameters(requestContents: OidcRequestContent, uri: Uri) {
-        if (uri.getQueryParameter(CLIENT_ID) != requestContents.clientId) {
-            throw ValidatorException("Request content does not match url parameters.")
-        }
-    }
-
-    internal fun deserializeJwsToken(serializedToken: String): JwsToken {
+    private fun deserializeJwsToken(serializedToken: String): JwsToken {
         return JwsToken.deserialize(serializedToken, serializer)
     }
 }
--- a/sdk/src/main/java/com/microsoft/did/sdk/credential/service/validators/OidcPresentationRequestValidator.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/credential/service/validators/OidcPresentationRequestValidator.kt
@@ -1,14 +1,12 @@
 package com.microsoft.did.sdk.credential.service.validators
 
-import android.net.Uri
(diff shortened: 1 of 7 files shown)


sdk/src/test/java/com/microsoft/did/sdk/credential/service/validators/JwtValidatorTest.kt
```diff
@@ -0,0 +1,114 @@
...
+        mockkObject(JwsToken)
+    }
+
+    private fun setUpResolver() {
+        every { mockedIdentifierDocument.publicKey } returns listOf(mockedIdentifierDocumentPublicKey)
+        every { mockedIdentifierDocumentPublicKey.toPublicKey() } returns mockedPublicKey
+    }
+
+    @Test
+    fun validateSignatureTest() {
```
AndiMiko, line 63: nit, write tests with Kotlin backtick method notation
e.g.
``valid signature is validated successfully``

sdk/src/main/java/com/microsoft/did/sdk/credential/service/validators/OidcPresentationRequestValidator.kt
```diff
@@ -18,38 +16,26 @@ import javax.inject.Singleton
...
     private fun getExpirationDeadlineInSeconds(expirationCheckTimeOffsetInMinutes: Int = 5): Long {
         val currentTimeInSeconds = Date().time / MILLISECONDS_IN_A_SECOND
         return currentTimeInSeconds + SECONDS_IN_A_MINUTE * expirationCheckTimeOffsetInMinutes
     }
-
-    private fun checkRequestParameters(requestContents: OidcRequestContent, uri: Uri) {
-        if (uri.getQueryParameter(CLIENT_ID) != requestContents.clientId) {
-            throw ValidatorException("Request content does not match url parameters.")
-        }
-    }
```
AndiMiko, line 50: So it was never necessary? Can you describe what the implications of this change are or why the client_id was there in the first place? (I think we briefly discussed it a while ago)

AndiMiko:
nit, name the PR better than "Bug fix" -> "Remove client_id from deeplink check"

sdk/src/main/java/com/microsoft/did/sdk/credential/service/validators/OidcPresentationRequestValidator.kt
symorton: Guillermo and I briefly talked a while back that the client_id in the qr code / deep link needs to match the client_id in the SIOP (presentation request). However, it was making qr codes too long. We just accepted that would be an issue. Last week, I talked to Danny and Brandon separately about this client_id parameter, and they both agree that this is not necessary, and a client_id in the SIOP is sufficient.

sdk/src/test/java/com/microsoft/did/sdk/credential/service/validators/JwtValidatorTest.kt
symorton: okay, I thought we changed our minds on this for some reason, but maybe that was just for instrumented tests?

symorton: updated names but I cannot put a space in between the words because IDE is giving me an "Identifier is not allowed in Android Projects error"

symorton: updated.

AndiMiko: yea in instrumented tests you can't use those kotlin methods for some reason

PR #107, opened 2020-08-03
Niganesh/pairwise key generation update

nithyaganeshng:
Updated pairwise key generation to use HMAC-SHA512.
Updated the test file to latest keys to verify if keys generated are deterministic

**Type of change:**
- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
@@ -21,7 +21,7 @@ import com.microsoft.did.sdk.util.Base64Url
 import com.microsoft.did.sdk.util.controlflow.PairwiseKeyException
 import com.microsoft.did.sdk.util.convertToBigEndian
 import com.microsoft.did.sdk.util.generatePublicKeyFromPrivateKey
-import com.microsoft.did.sdk.util.reduceKeySeedSize
+import com.microsoft.did.sdk.util.reduceKeySeedSizeAndConvertToUnsigned
 import com.microsoft.did.sdk.util.publicToXY
 import java.nio.ByteBuffer
 import java.util.Locale
@@ -38,13 +38,13 @@ class EllipticCurvePairwiseKey @Inject constructor() {
         val subtleCrypto: SubtleCrypto =
             crypto.subtleCryptoFactory.getMessageAuthenticationCodeSigners(W3cCryptoApiConstants.Hmac.value, SubtleCryptoScope.PRIVATE)
 
-        var pairwiseKeySeed = generatePairwiseSeed(subtleCrypto, masterKey, peerId)
-        pairwiseKeySeed = reduceKeySeedSize(pairwiseKeySeed)
+        val pairwiseKeySeedSigned = generatePairwiseSeed(subtleCrypto, masterKey, peerId)
+        val pairwiseKeySeedUnsigned = reduceKeySeedSizeAndConvertToUnsigned(pairwiseKeySeedSigned)
 
-        val pubKey = generatePublicKeyFromPrivateKey(pairwiseKeySeed)
+        val pubKey = generatePublicKeyFromPrivateKey(pairwiseKeySeedUnsigned)
         val xyData = publicToXY(pubKey)
 
-        val pairwiseKeySeedInBigEndian = convertToBigEndian(pairwiseKeySeed)
+        val pairwiseKeySeedInBigEndian = convertToBigEndian(pairwiseKeySeedUnsigned)
 
         return createPairwiseKeyFromPairwiseSeed(algorithm, pairwiseKeySeedInBigEndian, xyData)
     }
--- a/sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
@@ -51,10 +51,10 @@ class EllipticCurvePairwiseKey @Inject constructor() {
     private fun generatePairwiseSeed(subtleCrypto: SubtleCrypto, masterKey: ByteArray, peerId: String): ByteArray {
         // Generate the pairwise seed
         val alg =
-            Algorithm(name = W3cCryptoApiConstants.HmacSha256.value)
+            Algorithm(name = W3cCryptoApiConstants.HmacSha512.value)
         val signingKey = JsonWebKey(
             kty = KeyType.Octets.value,
-            alg = JoseConstants.Hs256.value,
+            alg = JoseConstants.Hs512.value,
             k = Base64Url.encode(masterKey)
         )
         val key = subtleCrypto.importKey(
--- a/sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
@@ -15,14 +15,13 @@ import com.microsoft.did.sdk.crypto.models.webCryptoApi.SubtleCrypto
 import com.microsoft.did.sdk.crypto.models.webCryptoApi.W3cCryptoApiConstants
 import com.microsoft.did.sdk.crypto.models.webCryptoApi.algorithms.Algorithm
 import com.microsoft.did.sdk.crypto.models.webCryptoApi.algorithms.EcKeyGenParams
-import com.microsoft.did.sdk.crypto.models.webCryptoApi.algorithms.EcdsaParams
 import com.microsoft.did.sdk.crypto.plugins.SubtleCryptoScope
 import com.microsoft.did.sdk.crypto.protocols.jose.JoseConstants
 import com.microsoft.did.sdk.util.Base64Url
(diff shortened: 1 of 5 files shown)


sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -35,11 +36,13 @@ class EllipticCurvePairwiseKey @Inject constructor() {
         val subtleCrypto: SubtleCrypto =
             crypto.subtleCryptoFactory.getMessageAuthenticationCodeSigners(W3cCryptoApiConstants.Hmac.value, SubtleCryptoScope.PRIVATE)
 
-        val pairwiseKeySeed = generatePairwiseSeed(subtleCrypto, masterKey, peerId)
+        var pairwiseKeySeed = generatePairwiseSeed(subtleCrypto, masterKey, peerId)
 
         if (supportedCurves.indexOf((algorithm as EcKeyGenParams).namedCurve.toUpperCase(Locale.ROOT)) == -1)
```
codeglobally, line 41: Move the if block before doing any actual work such as getting code signers and generating pairwise seeds. Prevents nugatory work.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -35,11 +36,13 @@ class EllipticCurvePairwiseKey @Inject constructor() {
...
             throw PairwiseKeyException("Curve ${algorithm.namedCurve} is not supported")
 
+        pairwiseKeySeed = getUnsignedModulus(pairwiseKeySeed)
```
codeglobally, line 44: This just feels wrong to me. Not clear to me from the function (no function comments) what the purpose of an unsigned modulus is and why the pairwiseSeed is being mutated

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
codeglobally, line 41: Move the check for supported curves before the generation of code signers and seeds. Avoids nugatory work and potential crypto battery drain.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -51,10 +54,10 @@ class EllipticCurvePairwiseKey @Inject constructor() {
     private fun generatePairwiseSeed(subtleCrypto: SubtleCrypto, masterKey: ByteArray, peerId: String): ByteArray {
         // Generate the pairwise seed
         val alg =
-            Algorithm(name = W3cCryptoApiConstants.HmacSha256.value)
+            Algorithm(name = W3cCryptoApiConstants.HmacSha512.value)
```
codeglobally, line 57: Can't add a comment in the createPairwiseKeyFromPairwiseSeed function below - but we have updated the use of SHA from 256 to 512, however this function still uses 256. Is this correct?

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: moved it.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: I was wondering how the tests passed and turns out algorithm in that method is not being used. Removed it.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: Renamed the method and also added comments. This method reduces seed generated using Sha512 to 256 bits.

AndiMiko: The comment hardly helped me to understand this method.

A BigInteger is initialized and then modulo is applied with some EC config? Is this behavior some well known cryptographic operation? If so, isn't it part of some library? If not, can you add a reference or descriptive explanation what this does? I don't see future devs understanding what this is doing.

Your comment may contain: Why this is done (why we need it and why it's done in this specific way)? What is done (you have that)? How it's done? Add references to some specification or something

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
```diff
@@ -32,13 +32,14 @@ class EllipticCurvePairwiseKey @Inject constructor() {
...
+
         val subtleCrypto: SubtleCrypto =
             crypto.subtleCryptoFactory.getMessageAuthenticationCodeSigners(W3cCryptoApiConstants.Hmac.value, SubtleCryptoScope.PRIVATE)
 
-        val pairwiseKeySeed = generatePairwiseSeed(subtleCrypto, masterKey, peerId)
-
-        if (supportedCurves.indexOf((algorithm as EcKeyGenParams).namedCurve.toUpperCase(Locale.ROOT)) == -1)
-            throw PairwiseKeyException("Curve ${algorithm.namedCurve} is not supported")
+        var pairwiseKeySeed = generatePairwiseSeed(subtleCrypto, masterKey, peerId)
+        pairwiseKeySeed = reduceKeySeedSize(pairwiseKeySeed)
```
AndiMiko, line 42: nit, since you moved this statements together in your last changes it's less of a code smell, but I would still prefer to use val instead of reassigning a byte array here. (use a new name for the new val)

Use `var` only for variables where the intent that they are supposed to change is clear (like `i` in a loop). Kotlin specifically allows for local val because using mutable variables is a common source of bugs after refactorings.

sdk/src/main/java/com/microsoft/did/sdk/util/EllipticCurveUtil.kt
```diff
@@ -64,4 +64,16 @@ fun generatePublicKeyFromPrivateKey(privateKey: ByteArray): ByteArray {
...
+
+/**
+ * Reduce size of pairwise key seed generated using SHA512 to 256 bits
+ */
+fun reduceKeySeedSize(keySeed: ByteArray): ByteArray {
+    val ecSpec = ECNamedCurveTable.getParameterSpec(Constants.SECP256K1_CURVE_NAME_EC)
+    var privateKey = BigInteger(1, keySeed).rem(ecSpec.n).toByteArray()
+
+    if (privateKey.size > 32)
+        privateKey = privateKey.sliceArray(1 until privateKey.size)
```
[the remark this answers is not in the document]
nithyaganeshng: It is either 32 or 33 bytes. If it is 33, first byte is 0 which is for sign. Yeah thought I will be overdoing it by refactoring it. But if it is not clear I will do it and also add name the function to indicate it is converting it to unsigned.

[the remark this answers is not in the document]
beejones: can't  you specify that the modulus operation is unsigned? A result off 33 bytes will return a different HMAC and as such typescript and android will not be compatible.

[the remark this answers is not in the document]
nithyaganeshng: This is done after HMAC. There is no unsigned modulus in kotlin so using rem to compute modulus. When it is converted to byte array if first byte is negative, 0 is prepended. So truncating the sign byte. It shouldn't affect other bytes and typescript SDK will be compatible with it.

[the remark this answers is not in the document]
nithyaganeshng: > It is either 32 or 33 bytes. If it is 33, first byte is 0 which is for sign. Yeah thought I will be overdoing it by refactoring it. But if it is not clear I will do it and also add name the function to indicate it is converting it to unsigned.

I realized this conversion was done in other places. Since this was third time, considered it a pattern and refactored it to a new method.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: I agree with general usage of val. But not sure here as it is the same content that is being converted. Created another variable anyway.

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/ellipticCurve/EllipticCurvePairwiseKey.kt
nithyaganeshng: Added detailed comment to the method.

PR #110, opened 2020-08-24
Niganesh/presentation exchange

nithyaganeshng:
Moved from attestation model to presentation exchange for presentation. These are the major changes included in this PR.
1. Replaced attestation model with presentation exchange input descriptors in presentation request
2. Added presentation exchange submission map along with attestation model in presentation response
3. Refactored Oidc Request/Response classes to have separate class for each service API and inherit common parameters from base class
4. Refactor Oidc response formatter to separate class for each service API
5. Remove nullable fields from protocol data model
6. Change tests impacted by above changes

**Solution:**
Used library A to complete scenario X with quality and style.

**Validation:**

**Type of change:**
- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:
- [x] High – Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)

**Work Item links**:

**Documentation Links**:


sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/oidc/OidcRequestContentForPresentation.kt
```diff
@@ -15,7 +15,7 @@ import kotlinx.serialization.Serializable
  * @see [OpenID Spec](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests)
  */
 @Serializable
-data class OidcRequestContent(
+data class OidcRequestContentForPresentation(
```
symorton, line 18: nit: This might be just me but I feel like instead of "ForPresentation" at the end of the name of the class, the class name could be "PresentationRequestContent". I think Presentation is one of the most important describing words for the class so it should be first.

symorton: oh I just saw PresentatioResponseContent was changed already too

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/oidc/PresentationResponseClaims.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Contents of an OpenID Self-Issued Token Response.
+ *
+ * @see [OpenID Spec](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests)
+ */
+@Serializable
+data class PresentationResponseClaims(
```
symorton, line 18: nice! So you figured out the serialization??

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/presentationexchange/CredentialPresentationInputDescriptors.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CredentialPresentationInputDescriptors(
+    var id: String = "",
+    val schema: Schema = Schema()
+) {
+    @SerialName("issuance")
```
symorton, line 13: I still dont fully understand the difference between making a value a param in the constructor vs a property in the class.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/ExchangeResponseFormatter.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val serializer: Serializer,
+    private val signer: TokenSigner
+) : OidcResponseFormatter {
+    fun formatExchangeResponse(exchangeRequest: ExchangeRequest, expiryInSeconds: Int): String {
+        val (iat, exp) = createIatAndExp(expiryInSeconds)
+        val jti = UUID.randomUUID().toString()
+        return createAndSignOidcResponseContentForExchange(exchangeRequest, iat, exp, jti)
+    }
+
+    private fun createAndSignOidcResponseContentForExchange(
```
symorton, line 30: nit: I don't believe you need ForExchange in this method name because the class has Exchange already in the class name

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/ExchangeResponseFormatter.kt
```diff
@@ -0,0 +1,54 @@
...
+            did = requester.id
+            publicKeyJwk = key.toJWK()
+            responseCreationTime = issuedTime
+            expirationTime = expiryTime
+            responseId = jti
+        }
+        return signContentsForExchange(contents, requester)
+    }
+
+    private fun signContentsForExchange(contents: ExchangeResponseClaims, responder: Identifier): String {
```
symorton, line 50: same here.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
```diff
@@ -0,0 +1,124 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class PresentationResponseFormatter @Inject constructor(
+    private val cryptoOperations: CryptoOperations,
+    private val serializer: Serializer,
+    private val verifiablePresentationFormatter: VerifiablePresentationFormatter,
+    private val signer: TokenSigner
+) : OidcResponseFormatter {
```
symorton, line 29: nit: Is there any shared logic between all of these Formatters that could be implemented an abstract class OidcResponseFormatter (instead of it being an interface)? Perhaps the createAndSignOidcResponseContentForPresentation logic?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/VerifiablePresentationFormatter.kt
```diff
@@ -37,9 +37,9 @@ class VerifiablePresentationFormatter @Inject constructor(
                 jti = jti,
                 vp = vp,
                 iss = did,
-                iat = iat,
-                nbf = iat,
-                exp = exp,
+                tokenIssuedTime = iat,
```
symorton, line 40: nit: iat, exp, nbf variables in formatter could be the better names as well :)

sdk/src/main/java/com/microsoft/did/sdk/crypto/keys/PublicKey.kt
```diff
@@ -53,7 +54,8 @@ abstract class PublicKey(val key: JsonWebKey) : IKeyStoreItem {
         val json = this.minimumAlphabeticJwk()
         val jsonUtf8 = stringToByteArray(json)
         val digest = crypto.subtleCryptoFactory.getMessageDigest(sha.name, SubtleCryptoScope.PUBLIC)
-        val hash = digest.digest(sha, jsonUtf8)
+        val md = MessageDigest.getInstance(Sha.SHA256.name)
```
symorton, line 57: nit: better name than md?

sdk/src/main/java/com/microsoft/did/sdk/crypto/models/webCryptoApi/W3cCryptoApiConstants.kt
```diff
@@ -50,7 +50,7 @@ enum class W3cCryptoApiConstants(val value: String) {
     Secp256r1("P-256"),
     Secp384r1("P-384"),
     Secp521r1("P-521"),
-    Secp256k1("P-256K"),
+    Secp256k1("secp256k1"),
```
symorton, line 53: great! this is the change ronny brought up?

sdk/src/main/java/com/microsoft/did/sdk/datasource/repository/VerifiableCredentialHolderRepository.kt
```diff
@@ -40,7 +45,9 @@ import javax.inject.Singleton
 class VerifiableCredentialHolderRepository @Inject constructor(
     database: SdkDatabase,
     private val apiProvider: ApiProvider,
-    private val formatter: OidcResponseFormatter,
+    @Named("issuanceResponseFormatter") private val issuanceResponseFormatter: OidcResponseFormatter,
```
symorton, line 48: nit: could we take in a mapping of OidcResponseFormatters?

symorton: that way if we need to create another formatter? We could just change the logic in Dagger by adding the new formatter to the mapping, instead of adding another parameter in this Repository? What do you thing @andimiko?

sdk/src/main/java/com/microsoft/did/sdk/datasource/repository/VerifiableCredentialHolderRepository.kt
```diff
@@ -85,18 +92,14 @@ class VerifiableCredentialHolderRepository @Inject constructor(
...
-        val formattedResponse = formatter.format(
-            responder = responder,
-            responseAudience = response.audience,
-            presentationsAudience = response.request.entityIdentifier,
+    suspend fun sendIssuanceResponse(
+        response: IssuanceResponse,
+        requestedVchMap: RequestedVchMap,
+        expiryInSeconds: Int = DEFAULT_EXPIRATION_IN_SECONDS
+    ): Result<VerifiableCredential> {
+        val formattedResponse = (issuanceResponseFormatter as IssuanceResponseFormatter).formatIssuanceResponse(
```
symorton, line 100: Why not pass in issuanceResponseFormatter as a IssuanceResponseFormatter as a parameter so no cast would be necessary? (if we dont use the mapping rec above)

sdk/src/main/java/com/microsoft/did/sdk/di/SdkModule.kt
```diff
@@ -121,4 +126,25 @@ internal class SdkModule {
     fun defaultValidator(validator: OidcPresentationRequestValidator): PresentationRequestValidator {
         return validator
     }
+
+    @Provides
+    @Singleton
+    @Named("issuanceResponseFormatter")
+    fun getIssuanceResponseFormatter(responseFormatter: IssuanceResponseFormatter): OidcResponseFormatter {
```
symorton, line 133: like here, maybe we could have a mapping for OidcResponseFormatters? Map<String, Formatter>?

sdk/src/main/java/com/microsoft/did/sdk/util/Constants.kt
```diff
@@ -9,9 +9,13 @@ object Constants {
 
     // OIDC Protocol Constants
     const val SELF_ISSUED = "https://self-issued.me"
-    const val SUB_JWK = "sub_jwk"
     const val PURE_ISSUANCE_FLOW_VALUE = "create"
 
+    //Presentation Exchange Constants
+    const val CREDENTIAL_PATH_IN_RESPONSE = "$.attestations.presentations"
```
symorton, line 15: will this be temporary until we get a JSON path lib in for interop in the future?

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -42,7 +42,7 @@ import javax.inject.Singleton
  */
 @Singleton
 class VerifiableCredentialManager @Inject constructor(
-    private val vchRepository: VerifiableCredentialHolderRepository,
+    val vchRepository: VerifiableCredentialHolderRepository,
```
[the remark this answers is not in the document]
nithyaganeshng: nice catch Sydney, it wasn't intentional. I was testing PE via SDK first and made it public to test those methods.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/VcServiceActionRequest.kt
```diff
@@ -10,7 +13,7 @@ import com.microsoft.did.sdk.identifier.models.Identifier
  */
 sealed class VcServiceActionRequest(val audience: String)
 
-class ExchangeRequest(val verifiableCredential: VerifiableCredential, val pairwiseDid: String) :
+class ExchangeRequest(val verifiableCredential: VerifiableCredential, val pairwiseDid: String, val requester: Identifier) :
```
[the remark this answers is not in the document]
nithyaganeshng: requester is the master DID of the user.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/oidc/OidcRequestContentForPresentation.kt
nithyaganeshng: Oh yes I agree with you. I missed it in Request class. A lot of name changes and refactoring happened :( I am going through them again to check if I have missed anymore.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/ExchangeResponseFormatter.kt
nithyaganeshng: I am removing all the "ForX" in my next commit. Thanks for pointing it out.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -104,43 +104,43 @@ class VerifiableCredentialManager @Inject constructor(
...
         return runResultTry {
             presentationRequestValidator.validate(request)
             Result.Success(Unit)
         }
     }
 
-    fun createIssuanceResponse(request: IssuanceRequest): IssuanceResponse {
-        return IssuanceResponse(request)
+    fun createIssuanceResponse(request: IssuanceRequest, responder: Identifier): IssuanceResponse {
+        return IssuanceResponse(request, responder)
```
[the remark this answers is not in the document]
nithyaganeshng: 😊

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -104,43 +104,43 @@ class VerifiableCredentialManager @Inject constructor(
      *
      * @param request to be validated.
      */
-    private suspend fun isRequestValid(request: PresentationRequest): Result<Unit> {
+    suspend fun isRequestValid(request: PresentationRequest): Result<Unit> {
```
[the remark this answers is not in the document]
nithyaganeshng: It was for testing within SDK. I made it private again.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -104,43 +104,43 @@ class VerifiableCredentialManager @Inject constructor(
...
         ): Result<VerifiableCredentialHolder> {
         return withContext(Dispatchers.IO) {
             runResultTry {
+                val responder = response.responder
                 val requestedVchMap = getExchangedVcs(
                     exchangeForPairwiseVerifiableCredential,
-                    response.getRequestedVchs(),
-                    responder).abortOnError()
-                val verifiableCredential = vchRepository.sendIssuanceResponse(response, requestedVchMap, responder, expiryInSeconds).abortOnError()
+                    responder,
```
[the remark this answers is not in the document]
nithyaganeshng: Moved it up to sending issuance/presentation response methods. It doesn't make a difference in function but is much clearer up here.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -187,9 +186,40 @@ class VerifiableCredentialManager @Inject constructor(
...
+                    it.value.cardId,
+                    vchRepository.getExchangedVerifiableCredential(it.component2(), responder).abortOnError(),
+                    it.value.owner,
+                    it.value.displayContract)
+            }
+            Result.Success(exchangedVcMap as RequestedVchMap)
+        }
+    }
+
+    suspend fun getExchangedVcs(
```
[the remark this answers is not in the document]
nithyaganeshng: Since issuance and presentation had different models it was different methods. I merged them into one method now and added a branching logic.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/oidc/PresentationResponseClaims.kt
nithyaganeshng: :)

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
nithyaganeshng: I thought there was shared logic when I created it. But on a closer look, it is not same across issuance/presentation and exchange. So removed inheritance now. This should make dependency injection much simpler now.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/VerifiablePresentationFormatter.kt
nithyaganeshng: You mean rename all the abbreviations to full names or vice versa?

sdk/src/main/java/com/microsoft/did/sdk/crypto/models/webCryptoApi/W3cCryptoApiConstants.kt
nithyaganeshng: yes this was the one. Oh this reminds me, I have to add typ to header. I will add it in next commit.

sdk/src/main/java/com/microsoft/did/sdk/datasource/repository/VerifiableCredentialHolderRepository.kt
nithyaganeshng: We don't need these anymore.

sdk/src/main/java/com/microsoft/did/sdk/di/SdkModule.kt
nithyaganeshng: We don't need it anymore :)

sdk/src/main/java/com/microsoft/did/sdk/util/Constants.kt
nithyaganeshng: Yes, when we do constraints we would need to add Json path lib in order to parse those. So just added it for now.

sdk/src/main/java/com/microsoft/did/sdk/datasource/repository/VerifiableCredentialHolderRepository.kt
nithyaganeshng: No casting needed now.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/presentationexchange/CredentialPresentationInputDescriptors.kt
nithyaganeshng: With default values passed we don't need the property method. In general property method is for optional values. But with default values anything in the constructor is optional. I am going to try with only constructor parameters.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -187,9 +186,40 @@ class VerifiableCredentialManager @Inject constructor(
...
     ): Result<RequestedVchMap> {
+        return runResultTry {
+            val exchangedVcMap = verifiableCredentialHolderRequestMappings.mapValues {
+                VerifiableCredentialHolder(
+                    it.value.cardId,
+                    vchRepository.getExchangedVerifiableCredential(it.component2(), responder).abortOnError(),
+                    it.value.owner,
+                    it.value.displayContract)
+            }
+            Result.Success(exchangedVcMap as RequestedVchMap)
```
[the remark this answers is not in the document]
nithyaganeshng: exchangedVcMap is Map which needs to be explicitly cast to its type alias.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/presentationexchange/CredentialPresentationInputDescriptors.kt
nithyaganeshng: Replaced all properties with constructors with default values.

sdk/src/main/java/com/microsoft/did/sdk/crypto/models/webCryptoApi/W3cCryptoApiConstants.kt
nithyaganeshng: I have added typ to JWT header.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -58,8 +58,8 @@ class VerifiableCredentialManager @Inject constructor(
                 val uri = verifyUri(stringUri)
                 val requestToken = getPresentationRequestToken(uri).abortOnError()
                 val tokenContents =
-                    serializer.parse(OidcRequestContent.serializer(), JwsToken.deserialize(requestToken, serializer).content())
-                val request = PresentationRequest(requestToken, tokenContents)
+                    serializer.parse(OidcRequestContentForPresentation.serializer(), JwsToken.deserialize(requestToken, serializer).content())
+                val request = PresentationRequest(requestToken, tokenContents, tokenContents.presentationDefinition)
```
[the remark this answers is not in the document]
nithyaganeshng: Having this as part of request is easier for wallet to consume and doesn't need to understand the structure of contract to retrieve it. But it is not needed it in the constructor as you said, so moved it inside the class and initialized it from contract object. Did the same for both issuance and presentation requests.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/FormatterHelpers.kt
```diff
@@ -5,17 +5,21 @@
...
     }
     val expiryInMilliseconds = 1000 * expiryInSeconds
     val expiration = currentTime + expiryInMilliseconds.toLong()
     val exp = floor(expiration / 1000f).toLong()
     return Pair(iat, exp)
-}
\ No newline at end of file
+}
+
+typealias RequestedVcIdToVchMap = Map<Pair<String, Int>, VerifiableCredentialHolder>
```
AndiMiko, line 25: The name suggests it's mapping a VcId to Vch, but the type is oddly a `Pair<String, Int>`. Looking in the code the Pair is created with something different then the VcId? `Pair(key.credentialType, key.validityInterval)`. Confusing.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
```diff
@@ -0,0 +1,124 @@
...
+    private val serializer: Serializer,
+    private val verifiablePresentationFormatter: VerifiablePresentationFormatter,
+    private val signer: TokenSigner
+) {
+    fun formatResponse(
+        requestedVchPresentationSubmissionMap: RequestedVchPresentationSubmissionMap = mutableMapOf(),
+        presentationResponse: PresentationResponse,
+        expiryInSeconds: Int = Constants.DEFAULT_EXPIRATION_IN_SECONDS
+    ): String {
+        val (iat, exp) = createIatAndExp(expiryInSeconds)
```
AndiMiko, line 35: nit, I know you copied this code from the other Formatter but please make an effort to improve this code by using proper names (as you already did on the data models as well)

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
```diff
@@ -0,0 +1,124 @@
...
+    ): Map<String, String> {
+        return requestedVcIdToVchMap.map { (key, value) ->
+            key.first to verifiablePresentationFormatter.createPresentation(
+                value.verifiableCredential,
+                key.second,
+                audience,
+                responder
+            )
+        }.toMap()
+    }
```
AndiMiko, line 122: It's hard for me to understand what these transformations are doing. Seems to me that it stems from what I commented earlier on that `RequestedVcIdToVchMap` is a weird type. Can we make this more straight forward?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
```diff
@@ -0,0 +1,124 @@
...
+        val attestationResponse = this.createAttestationClaimModel(
+            requestedVchPresentationSubmissionMap,
+            presentationResponse.request.entityIdentifier,
+            presentationResponse.responder
+        )
+        val credentialPresentationSubmissionDescriptors =
+            presentationResponse.getRequestedVchClaims().map {
+                PresentationSubmissionDescriptor(
+                    it.component1().id,
+                    "${Constants.CREDENTIAL_PATH_IN_RESPONSE}.${it.component1().id}",
```
AndiMiko, line 46: nit, key instead of component1

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
```diff
@@ -0,0 +1,124 @@
...
+            }
+        val credentialPresentationSubmission = PresentationSubmission(credentialPresentationSubmissionDescriptors)
+        return createAndSignOidcResponseContent(
+            presentationResponse,
+            iat,
+            exp,
+            jti,
+            attestationResponse,
+            credentialPresentationSubmission
+        )
```
AndiMiko, line 59: nit, this class in general is bugging me due to the amount of parameters passed. Makes it hard to read. Not sure if this can be improved, maybe you have an idea.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/VerifiablePresentationFormatter.kt
AndiMiko: to full names :)

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/VcServiceActionRequest.kt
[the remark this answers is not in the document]
nithyaganeshng: It was passed in as a separate argument to response formatting method earlier. Now it is all included in the request object itself which is then passed to format method.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -111,36 +114,35 @@ class VerifiableCredentialManager @Inject constructor(
...
-                val requestedVchMap = getExchangedVcs(
-                    exchangeForPairwiseVerifiableCredential,
-                    response.getRequestedVchs(),
-                    responder).abortOnError()
-                val verifiableCredential = vchRepository.sendIssuanceResponse(response, requestedVchMap, responder, expiryInSeconds).abortOnError()
+                val requestedVchMap = if (exchangeForPairwiseVerifiableCredential)
+                    getExchangedVcs(response).abortOnError()
+                else
+                    response.getRequestedVchs()
+                //TODO: How to avoid this suppression
```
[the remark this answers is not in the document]
nithyaganeshng: I actually combined thinking it was basically doing the same thing. I will revert it back to 2 methods if that is clearer.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
```diff
@@ -5,18 +5,21 @@
...
 @Serializable
-sealed class Request(val attestations: CredentialAttestations, val entityName: String = "", val entityIdentifier: String = "")
+sealed class Request(val entityName: String = "", val entityIdentifier: String = "")
 
 @Serializable
 class IssuanceRequest(val contract: VerifiableCredentialContract, val contractUrl: String) :
-    Request(contract.input.attestations, contract.display.card.issuedBy, contract.input.issuer)
+    Request(contract.display.card.issuedBy, contract.input.issuer) {
+    val attestations = contract.input.attestations
+}
```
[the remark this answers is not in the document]
nithyaganeshng: It populates the field from what is already there. But without explicitly having this field, wallet should parse the whole path when it just wants attestations and should be aware of the structure.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Response.kt
```diff
@@ -10,19 +10,18 @@ import com.microsoft.did.sdk.credential.service.models.attestations.Presentation
...
  *
  * @param audience entity to send the response to.
  */
-sealed class Response(open val request: Request, val audience: String) {
-    
-    private val requestedVchMap: RequestedVchMap = mutableMapOf()
-
+sealed class Response(open val request: Request, val audience: String, open val responder: Identifier)
+class IssuanceResponse(override val request: IssuanceRequest, override val responder: Identifier) : Response(request, request.contract.input.credentialIssuer, responder) {
+    private var requestedVchMap: RequestedVchMap = mutableMapOf()
```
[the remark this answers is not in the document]
nithyaganeshng: I knew that :) but do we want to make them public? If we make it public, it could be cleared as well. I thought that was the reason we specifically made it private and only have methods to set and get. Isn't it the reason?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Response.kt
```diff
@@ -48,10 +47,24 @@ sealed class Response(open val request: Request, val audience: String) {
...
+    }
+
+    fun getRequestedVchClaims(): RequestedVchPresentationSubmissionMap {
+        return requestedVchPresentationSubmissionMap
+    }
 
     fun createReceiptsForPresentedVerifiableCredentials(entityDid: String, entityName: String): List<Receipt> {
         val receiptList = mutableListOf<Receipt>()
-        requestedVchMap.forEach {
+        requestedVchPresentationSubmissionMap.forEach {
```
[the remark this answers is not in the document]
nithyaganeshng: I know, I was so tempted to remove these but I have learnt by now that it affects a lot of tests and I don't want to re-do them :( After presentation exchange my goal is to get that PR in.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Response.kt
```diff
@@ -10,19 +10,18 @@ import com.microsoft.did.sdk.credential.service.models.attestations.Presentation
...
  * Response formed from a Request.
  *
  * @param audience entity to send the response to.
  */
-sealed class Response(open val request: Request, val audience: String) {
-    
-    private val requestedVchMap: RequestedVchMap = mutableMapOf()
-
+sealed class Response(open val request: Request, val audience: String, open val responder: Identifier)
+class IssuanceResponse(override val request: IssuanceRequest, override val responder: Identifier) : Response(request, request.contract.input.credentialIssuer, responder) {
```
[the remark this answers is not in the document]
nithyaganeshng: If we pass them without overriding, it is interpreted as Request (base) object  since only that is defined and it needs explicit casting when it is used as issuanceResponse.request to IssuanceRequest.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/FormatterHelpers.kt
nithyaganeshng: It is id of a VC in presentation exchange and credential type in attestation model :( This map is trying to bridge the gap and act as a common holder. Will think of a better name. Please provide any suggestions if you have.

nithyaganeshng: I mean first element in pair is either of those models and second is always validity interval.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
nithyaganeshng: When it was one formatter class there were 2 methods each for issuance and presentation. Felt like lot of duplicate code and refactored. Now they have their own formatter classes, when I think about it this is an overkill. I will just go back and have separate methods.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
nithyaganeshng: Although public facing method takes few parameters, private methods do take a lot of parameters. Oh instead of passing everything to a method, I will just create an response object inside this method. I will refactor this method differently.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -150,59 +152,66 @@ class VerifiableCredentialManager @Inject constructor(
...
             }
         }
     }
 
-    private suspend fun getExchangedVcs(
-        exchangeForPairwiseVerifiableCredential: Boolean,
-        requestedVchMap: RequestedVchMap,
-        responder: Identifier
-    ): Result<RequestedVchMap> {
+    private suspend fun getExchangedVcs(response: Response): Result<MutableMap<out Any, VerifiableCredentialHolder>> {
```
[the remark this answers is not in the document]
nithyaganeshng: I removed this method. But private method for exchanging VC still uses runResultTry as in invokes network call via repository and it should be handled in there.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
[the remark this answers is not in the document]
nithyaganeshng: Reverted back to separate methods for issuance and presentation.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialManager.kt
```diff
@@ -150,59 +152,66 @@ class VerifiableCredentialManager @Inject constructor(
...
-                        vchRepository.getExchangedVerifiableCredential(it.component2(), responder).abortOnError(),
-                        it.value.owner,
-                        it.value.displayContract)
+            when (response) {
+                is IssuanceResponse -> {
+                    val verifiableCredentialHolderRequestMappings = response.getRequestedVchs()
+                    val exchangedVcMap = verifiableCredentialHolderRequestMappings.mapValues {
+                        VerifiableCredentialHolder(
+                            it.value.cardId,
+                            vchRepository.getExchangedVerifiableCredential(it.component2(), response.responder).abortOnError(),
```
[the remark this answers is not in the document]
nithyaganeshng: made this change wherever it was used.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/FormatterHelpers.kt
nithyaganeshng: Removed usage of this map and added separate methods for issuance and presentation.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
nithyaganeshng: done

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/FormatterHelpers.kt
```diff
@@ -5,17 +5,21 @@
...
 import java.util.Date
 import kotlin.math.floor
 
-fun createIatAndExp(expiryInSeconds: Int): Pair<Long, Long?> {
+fun createIatAndExp(expiryInSeconds: Int): Pair<Long, Long> {
     val currentTime = Date().time
     val iat = floor(currentTime / 1000f).toLong()
     if (expiryInSeconds == -1) {
-        return Pair(iat, null)
+        throw FormatterException("Expiry for OIDC Responses cannot be null")
```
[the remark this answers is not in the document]
nithyaganeshng: Changed the exception message. I will check what -1 means.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Response.kt
```diff
@@ -48,10 +47,24 @@ sealed class Response(open val request: Request, val audience: String) {
...
+    fun addRequestedVchClaims(
+        credentialPresentationInputDescriptor: CredentialPresentationInputDescriptor,
+        vch: VerifiableCredentialHolder
+    ) {
+        requestedVchPresentationSubmissionMap[credentialPresentationInputDescriptor] = vch
+    }
+
+    fun getRequestedVchClaims(): RequestedVchPresentationSubmissionMap {
+        return requestedVchPresentationSubmissionMap
+    }
```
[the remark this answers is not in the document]
nithyaganeshng: I have commented in PresentationResponse class.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
nithyaganeshng: We were passing all the values to create an object and do something else as well. Refactored it create object in the same method. Now we don't pass everything around.

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/credentialOperations/FetchContractNetworkOperation.kt
```diff
@@ -8,8 +8,17 @@ package com.microsoft.did.sdk.datasource.network.credentialOperations
...
+import com.microsoft.did.sdk.util.controlflow.IssuanceException
+import com.microsoft.did.sdk.util.controlflow.Result
 import retrofit2.Response
 
-class FetchContractNetworkOperation(val url: String, apiProvider: ApiProvider) : GetNetworkOperation<VerifiableCredentialContract, VerifiableCredentialContract>() {
+class FetchContractNetworkOperation(val url: String, apiProvider: ApiProvider) :
+    GetNetworkOperation<VerifiableCredentialContract, VerifiableCredentialContract>() {
     override val call: suspend () -> Response<VerifiableCredentialContract> = { apiProvider.issuanceApis.getContract(url) }
+
+    override fun onSuccess(response: Response<VerifiableCredentialContract>): Result<VerifiableCredentialContract> {
```
nithyaganeshng, line 19: claims in self issued attestations have a type but there is no type in idtoken attestations. This was causing failures if type is made non-nullable. This parses the response or contract and sets type to empty string if it is not present. We could now make type non-nullable. I am also checking with service team if there is a reason for this. This failed because serialization library was expecting the field to be non-null when it was null in fact. If we have few such cases, we can consider writing custom serializer to convert null in json to empty strings. @AndiMiko @symorton  what do you think about it?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
```diff
@@ -5,18 +5,21 @@
...
 
 @Serializable
-class PresentationRequest(val serializedToken: String, val content: OidcRequestContent) :
-    Request(content.attestations, content.registration.clientName, content.iss)
\ No newline at end of file
+class PresentationRequest(val serializedToken: String, val requestContent: PresentationRequestContent) :
+    Request(requestContent.registration.clientName, requestContent.iss) {
+    val presentationDefinition = requestContent.presentationDefinition
+}
```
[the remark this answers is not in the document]
nithyaganeshng: Removed these and added a method for easy access of these fields in both requests.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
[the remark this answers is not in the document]
nithyaganeshng: Removed them.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Response.kt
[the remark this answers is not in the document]
nithyaganeshng: Keeping the override as per our discussion.

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/credentialOperations/FetchContractNetworkOperation.kt
nithyaganeshng: I debugged it further and found out that default values are assigned when kotlin serialization is used. While making network calls, retrofit uses gson which is causing this issue and assigning null instead of default. I am still figuring out a way to get around this is gson and retrofit. But I don't think it is a blocker for presentation exchange pr.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/FormatterHelpers.kt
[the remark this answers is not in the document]
AndiMiko: The new message is also not perfectly matching what the code says. The code is only excluding "-1", but your message says can't be negative. That's unlikely to be dramatic, but the fact that you now throwing for -1 at all seems like a bug to me. Either the code before didn't make sense at all or it actually covered a valid scenario "-1" that represented something like "no expiration date set" which would crash on cards where the expiration isn't set. This might be a dangerous bug since it may surface only much much later when a contract uses no expiration set, which might not be present in our test scenarios.

Can you confirm now please?

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/credentialOperations/FetchContractNetworkOperation.kt
AndiMiko: Makes sense since Gson is not a Kotlin-first library it seems not to be aware of Kotlin default parameters. Please come back to this asap as I see this behavior causing more issues in the future (since devs obviously would expect the defaults to be set).

(create a ADO task)

nithyaganeshng:
> Please follow up on the "-1" behavior of FormatterHelpers. Everything else looks good to go :)

Since we don't actually validate the expiration and use default values, removed it now. When we validate the expiration, we can come up with a strategy for it.

nithyaganeshng:
I have added few more changes to this PR for upgrading kotlin serialization version to 1.0.0-RC and in turn kotlin version to 1.4.0. There were a few changes in API so modified our classes accordingly.

AndiMiko (approved):
Since there is no direct relation to PE I'd usually prefer to make that Kotlin upgrade in a separate PR. Makes it easier to investigate later and roll back in case there are issues. But that's fine now :)

nithyaganeshng:
> Since there is no direct relation to PE I'd usually prefer to make that Kotlin upgrade in a separate PR. Makes it easier to investigate later and roll back in case there are issues. But that's fine now :)

I might have to roll back the changes already :( Everything works fine on local tests but having issues in pipeline.

symorton (approved):

PR #111, opened 2020-08-26
Improve Exceptions

AndiMiko:


sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
```diff
@@ -5,54 +5,56 @@
...
+class ResolverException(message: String, cause: Throwable? = null) : SdkException(message, cause)
 
-class RegistrarException(message: String? = null, cause: Throwable? = null) : SdkException(message, cause)
+class RegistrarException(message: String, cause: Throwable? = null) : SdkException(message, cause)
 
-open class NetworkException(message: String? = null, cause: Throwable? = null) : SdkException(message, cause)
+open class LocalNetworkException(message: String, cause: Throwable? = null) : SdkException(message, cause, true)
 
-class ServiceUnreachableException(message: String? = null, cause: Throwable? = null) : NetworkException(message, cause)
+open class NetworkException(val requestId: String, message: String, retryable: Boolean) : SdkException(message, null, retryable)
```
nithyaganeshng, line 52: Just for my understanding, localnetworkexception happens if phone loses connectivity and networkexception happens on the service? If so what is service unreachable then? Also wouldn't all these three be retryable by default?

AndiMiko: That's correct.

I didn't put default values as I want the networklayer to make concious decisions about it's retryablity. As you see in BaseNetworkOperation all 4xx are actually not retryable right now. Service unreachable 5xx is retryable.

(This needs more attention though in the future! I made some assumptions here for now)

nithyaganeshng: 👍

PR #112, opened 2020-09-05
Add delete and archive feature

AndiMiko:


sdk/src/main/java/com/microsoft/did/sdk/datasource/db/SdkDatabase.kt
```diff
@@ -24,15 +22,13 @@ import com.microsoft.did.sdk.identifier.models.Identifier
  * More info:
  * https://developer.android.com/topic/libraries/architecture/room
  */
-@Database(entities = [VerifiableCredential::class, VerifiableCredentialHolder::class, Identifier::class, Receipt::class], version = 2)
```
nithyaganeshng, line 27: We don't have two tables anymore for VerifiableCredential and its Holder?

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
```diff
@@ -50,8 +50,20 @@ abstract class BaseNetworkOperation<S, T> {
         val requestId = response.headers()["request-id"] ?: "?"
         return when (response.code()) {
             401 -> Result.Failure(UnauthorizedException(requestId, "${response.code()}: ${response.errorBody()?.string()}", false))
-            402, 403, 404 -> Result.Failure(ServiceErrorException(requestId, "${response.code()}: ${response.errorBody()?.string()}", false))
-            500, 501, 502, 503 -> Result.Failure(ServiceUnreachableException(requestId, "${response.code()}: ${response.errorBody()?.string()}", true))
+            402, 403, 404 -> Result.Failure(
```
nithyaganeshng, line 53: It is not related to changes for delete. But I just noticed we don't check for 400 here.

sdk/src/main/java/com/microsoft/did/sdk/datasource/repository/VerifiableCredentialHolderRepository.kt
```diff
@@ -178,7 +177,6 @@ class VerifiableCredentialHolderRepository @Inject constructor(
                     result.payload,
                     request.verifiableCredential.picId
                 )
-                this.insert(verifiableCredential)
```
nithyaganeshng, line 181: oh so instead of saving every exchanged VC in a table here, it is returned here and embedded as objects in vch table in layer above(manager)?

nithyaganeshng:
Some of the unit and instrumented tests are failing because of removed VerifiableCredentialDao class. Could you please fix those as part of this PR?

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/DbMigrations.kt
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved
+
+package com.microsoft.did.sdk.datasource.db
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+object DbMigrations {
```
symorton, line 8: does this mean that customers will not have to reinstall the app this time?

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/dao/VerifiableCredentialHolderDao.kt
```diff
@@ -10,23 +10,30 @@ import androidx.room.Dao
...
 
-    @Query("SELECT * FROM VerifiableCredentialHolder")
-    fun queryAllVcs(): List<VerifiableCredentialHolder>
+    @Query("SELECT * FROM VerifiableCredentialHolder WHERE isArchived = 0")
+    fun queryAllActiveVchs(): List<VerifiableCredentialHolder>
 
-    @Query("SELECT * FROM VerifiableCredentialHolder where picId = :id")
-    fun getVcById(id: String): LiveData<VerifiableCredentialHolder>
+    @Query("SELECT * FROM VerifiableCredentialHolder WHERE isArchived = 1")
+    fun getArchivedVchs(): LiveData<List<VerifiableCredentialHolder>>
```
symorton, line 26: we only need a LiveData version of this call?

symorton (approved):

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/SdkDatabase.kt
AndiMiko: We haven't. The VC table was only for pairwise caching, which I removed. It causes trouble to have the same data stored in two tables because syncing them is not guaranteed.

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
AndiMiko: Yes, we don't. It false in the else tree. There are also much more 4xx codes we don't check. We should rework this to be for example 400 =< x < 500 and then even further exclude some codes in detail. This should be part of the ongoing debugability work.

sdk/src/main/java/com/microsoft/did/sdk/datasource/repository/VerifiableCredentialHolderRepository.kt
AndiMiko: exchanged pairwise vcs are not saved at all anymore. This caching adds complexity and trouble down the road for little to no benefit (it's an uncommon scenario that you present the same exact VC to the same RP twice)

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/DbMigrations.kt
AndiMiko: yes

sdk/src/main/java/com/microsoft/did/sdk/datasource/db/dao/VerifiableCredentialHolderDao.kt
AndiMiko: For now yes, but we might discuss to always add both versions for consistency. (though we won't use it for now)

PR #115, opened 2020-09-18
Fix database migration and add constant

AndiMiko:


sdk/src/main/java/com/microsoft/did/sdk/util/Constants.kt
```diff
@@ -20,6 +20,7 @@ object Constants {
     const val CONTEXT = "@context"
     const val VP_CONTEXT_URL = "https://www.w3.org/2018/credentials/v1"
     const val VERIFIABLE_PRESENTATION_TYPE = "VerifiablePresentation"
+    const val VERIFIABLE_CREDENTIAL_DEFAULT_TYPE = "VerifiableCredential"
```
nithyaganeshng, line 23: Is this being used anywhere?

AndiMiko: It's used in Authenticator. Since it's a constant of the VC spec I have put it in the SDK even though the SDK doesn't make use of it as of yet

nithyaganeshng: 👍

nithyaganeshng (approved):

PR #119, opened 2020-09-25
Standardized version numbers in build.gradle

AndiMiko:
Standardized version numbers while refactoring to own module

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -20,6 +20,11 @@ android.enableJetifier=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
 
+build_tools_version=30.0.2
+compile_sdk_version=30
+min_sdk_version=23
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -20,6 +20,11 @@ android.enableJetifier=true
 # Kotlin code style for this project: "official" or "obsolete":
 kotlin.code.style=official
 
+build_tools_version=30.0.2
+compile_sdk_version=30
```
nithyaganeshng, line 24: Just curious, does bumping up compiler sdk version have any advantage for us?

AndiMiko: it doesn't directly, but I aligned it with Authenticator just in case :)

nithyaganeshng: 👍

nithyaganeshng (approved):

PR #121, opened 2020-10-08
Niganesh/spongycastle signature bug

nithyaganeshng:
**Problem:**
Signature created by Spongycastle were failing sometimes during validation by RP.

**Solution:**
Signature generated sometimes had less than 32 bytes which needed explicit padding in those cases which was added.

**Validation:**
Added a test to generate quite a few signatures and verified it was the right size.

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:


sdk/src/test/java/com/microsoft/did/sdk/crypto/plugins/SpongyCastleTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    private fun generatePublicKeyFromPrivateKey(privateKey: ByteArray): ByteArray {
+        val keyFactory = KeyFactory.getInstance(AndroidConstants.Ec.value)
+        val ecSpec = ECNamedCurveTable.getParameterSpec(Constants.SECP256K1_CURVE_NAME_EC)
+        val ecPoint = ecSpec.g.multiply(BigInteger(1, privateKey))
+        val pubKeySpec = ECPublicKeySpec(ecPoint, ecSpec)
+        val publicKey = keyFactory.generatePublic(pubKeySpec)
+        return byteArrayOf(Secp256k1Provider.Secp256k1Tag.UNCOMPRESSED.byte) + (publicKey as BCECPublicKey).q.normalize().xCoord.encoded +
+            publicKey.q.normalize().yCoord.encoded
+    }
+}
```
AndiMiko, line 92: Thanks for adding those thorough tests 👍👍

sdk/src/test/java/com/microsoft/did/sdk/crypto/plugins/SpongyCastleTest.kt
```diff
@@ -0,0 +1,92 @@
...
+        val ecParams = ECNamedCurveTable.getParameterSpec(Constants.SECP256K1_CURVE_NAME_EC)
+        return ECDomainParameters(ecParams.curve, ecParams.g, ecParams.n, ecParams.h)
+    }
+
+    @Test
+    fun `generateSameSizeSignTest`() {
+        val privateKey = generatePrivateKey()
+        val publicKey = generatePublicKeyFromPrivateKey(privateKey)
+        for(i in 1..1000) {
+            val data = Random.nextBytes(1000)
```
AndiMiko, line 48: using random in tests is problematic because if the test fails and you rerun the test you a) won't be able to see what the issue was (no repro) b) you can rerun until it's green and ignore the failure 

can you just generate these arrays once and hardcode them here?

AndiMiko (approved):

sdk/src/test/java/com/microsoft/did/sdk/crypto/plugins/SpongyCastleTest.kt
nithyaganeshng: The idea of this test is to generate random payload and sign to check if it sometimes creates signature lesser than 32 bytes. It shouldn't fail whatever the random value is. If it fails, we need to look into our signing method then.

nithyaganeshng: I can have some static text and append numbers to it and test. I think that should work.

nithyaganeshng: Removed random and made the change.

PR #125, opened 2020-10-16
Niganesh/send user agent in header

nithyaganeshng:
User-Agent will now be sent in the request header for all the requests.

**Validation:**
Verified that the request header comes back as part of response.

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
@@ -43,7 +43,7 @@ object VerifiableCredentialSdk {
      * Initializes VerifiableCredentialSdk
      *
      * @param context context instance
-     * @param userAgentInfo it contains name and version of the client
+     * @param userAgentInfo it contains name and version of the client. It will be used in User-Agent header for all the requests.
      * @param logConsumer logger implementation to be used
      * @param registrationUrl url used to register DID
      * @param resolverUrl url used to resolve DID
--- a/sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
@@ -44,13 +44,13 @@ object VerifiableCredentialSdk {
     fun init(
         context: Context,
         logConsumer: SdkLog.Consumer = DefaultLogConsumer(),
-        walletInfo: String,
+        userAgentInfo: String,
         registrationUrl: String = "",
         resolverUrl: String = "https://beta.discover.did.microsoft.com/1.0/identifiers"
     ) {
         val sdkComponent = DaggerSdkComponent.builder()
             .context(context)
-            .walletInfo(walletInfo)
+            .userAgentInfo(userAgentInfo)
             .registrationUrl(registrationUrl)
             .resolverUrl(resolverUrl)
             .build()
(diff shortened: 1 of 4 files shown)


sdk/src/main/java/com/microsoft/did/sdk/di/UserAgentInterceptor.kt
```diff
@@ -0,0 +1,20 @@
...
+package com.microsoft.did.sdk.di
+
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.io.IOException
+
+class UserAgentInterceptor(private val userAgentInfo: String) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Interceptor.Chain): Response {
```
nithyaganeshng, line 12: Although we can add this while building the okhttpclient object, felt it was cluttering up the method. So added a separate class for it.

sdk/src/main/java/com/microsoft/did/sdk/di/SdkComponent.kt
```diff
@@ -47,5 +47,8 @@ internal interface SdkComponent {
 
         @BindsInstance
         fun registrationUrl(@Named("registrationUrl") registrationUrl: String): Builder
+
+        @BindsInstance
+        fun userAgentInfo(@Named("userAgentInfo") userAgentInfo: String): Builder
```
symorton, line 52: nit: Naming, why not just userAgentVersion?

nithyaganeshng: I intentionally changed it because we append the app name before the version like the sample in ado.

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
```diff
@@ -44,11 +44,13 @@ object VerifiableCredentialSdk {
     fun init(
         context: Context,
         logConsumer: SdkLog.Consumer = DefaultLogConsumer(),
+        userAgentInfo: String,
```
AndiMiko, line 47: Do we want to have a default value? Maybe it's ok to force the client to add a userAgentInfo. Can you add a comment to the init method to describe what userAgentInfo is supposed to contain (Clientname + version. I guess version may be optional) and where it gets used (in the header).

Also if it hasn't a default value please put it above `logConsumer`, such that the non-default params can be supplied without forcing to either supply a defaulted parameter or using named params.

sdk/src/main/java/com/microsoft/did/sdk/di/UserAgentInterceptor.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Chain): Response {
+        val originalRequest = chain.request()
+        val requestWithUserAgentInfo = originalRequest.newBuilder()
+            .header("User-Agent", userAgentInfo)
+            .build()
+        return chain.proceed(requestWithUserAgentInfo)
+    }
+
```
AndiMiko, line 20: nit whitespace

AndiMiko (approved):

sdk/src/main/java/com/microsoft/did/sdk/VerifiableCredentialSdk.kt
nithyaganeshng: Yes, I think its ok too to force the client to send the information. At least client name would do even if version is not sent. I moved userAgentInfo above logConsumer.

PR #126, opened 2020-10-21
Migrate to Kotlin 1.4.10 and stable Kotlin Serialization 1.0.0

AndiMiko:
I will need to use serialization quite a bit for backup & restore and I figured we need to update to this stable version anyhow sooner or later. Also Kotlin 1.4+ is required and Authenticator team has also expressed the desire to upgrade.

**Type of change:**
- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [X] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:
- [X] High – Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)


sdk/src/test/java/com/microsoft/did/sdk/util/SerializationTest.kt
```diff
@@ -5,7 +5,6 @@ import com.microsoft.did.sdk.credential.service.models.serviceResponses.Revocati
 import com.microsoft.did.sdk.crypto.models.webCryptoApi.JsonWebKey
 import com.microsoft.did.sdk.identifier.models.identifierdocument.IdentifierDocument
 import com.microsoft.did.sdk.identifier.models.identifierdocument.IdentifierDocumentPublicKey
-import com.microsoft.did.sdk.util.serializer.Serializer
```
nithyaganeshng, line 8: This class has some tests for serialization using Serializer class. I am glad we don't need Serializer class anymore. We could remove this test class as well. If we want to test serialization, we can probably use Json directly.

sdk/src/test/java/com/microsoft/did/sdk/identifier/resolvers/ResolverTest.kt
```diff
@@ -8,7 +8,6 @@ import com.microsoft.did.sdk.util.controlflow.LocalNetworkException
 import com.microsoft.did.sdk.util.controlflow.ResolverException
 import com.microsoft.did.sdk.util.controlflow.Result
 import com.microsoft.did.sdk.util.controlflow.ServiceErrorException
-import com.microsoft.did.sdk.util.serializer.Serializer
```
nithyaganeshng, line 11: For all the tests where this import was removed, we can use Json directly

nithyaganeshng (changes requested):
Changes look good to me. Could you please fix the tests? Also, could you please add the dependency in cg?

sdk/src/test/java/com/microsoft/did/sdk/identifier/resolvers/ResolverTest.kt
AndiMiko: (can't respond on the other comment of yours)

Oh yea. I will verify that tests still work, forgot to verify. Will also update version in cg

AndiMiko:
@nithyaganeshng need a rereview :)

sdk/src/androidTest/java/com/microsoft/did/sdk/Serializer.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved
+
+package com.microsoft.did.sdk
+
+import com.microsoft.did.sdk.di.SdkModule
+
+// Keep in sync with `fun defaultJsonSerializer()` in SdkModule
```
nithyaganeshng, line 7: nit, this variable is always in sync with SdkModule now. we can remove this?

sdk/src/test/java/com/microsoft/did/sdk/util/Serializer.kt
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved
+
+package com.microsoft.did.sdk.util
+
+import com.microsoft.did.sdk.di.SdkModule
+
+// Keep in sync with `fun defaultJsonSerializer()` in SdkModule
```
nithyaganeshng, line 7: Same comment as above. Just realized that we are duplicating the same class in both unit and instrumented test. We removed common folder structure to reuse amongst all the tests, right? May be if we have more occurrences we can add it back.

nithyaganeshng (approved):

sdk/src/androidTest/java/com/microsoft/did/sdk/Serializer.kt
AndiMiko: you are correct, will remove. Good example of a bad comment, introduced and invalidated the same comment within the same PR (it made sense in a commit before) 😅😅😅

sdk/src/test/java/com/microsoft/did/sdk/util/Serializer.kt
AndiMiko: Wasn't aware we had a common one before. I have no hard feelings either way

nithyaganeshng: yeah it got removed as part of the same PR :) But I think one instance is not a pattern. We can refactor later if we see a pattern.

PR #134, opened 2020-11-25
Test performance of sign/verify primitives

beejones:
**Problem:**
Adding performance tests for the crypto primitives

**Solution:**
Add special test so Logcat can show the performance with filter: PerfTest->

**Validation:**
New unit tests

**Type of change:**
- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [x] Test
- [ ] Logging/Telemetry

**Risk**:

### DIFF
--- a/sdk/src/androidTest/java/com/microsoft/did/sdk/crypto/protectors/TokenSignerPerformanceTest.kt
+++ b/sdk/src/androidTest/java/com/microsoft/did/sdk/crypto/protectors/TokenSignerPerformanceTest.kt
@@ -9,24 +9,38 @@ import android.content.Context
 import androidx.test.internal.runner.junit4.AndroidJUnit4ClassRunner
 import androidx.test.platform.app.InstrumentationRegistry
 import com.microsoft.did.sdk.credential.service.protectors.TokenSigner
+import com.microsoft.did.sdk.credential.service.validators.JwtValidator
 import com.microsoft.did.sdk.crypto.CryptoOperations
 import com.microsoft.did.sdk.crypto.keyStore.AndroidKeyStore
+import com.microsoft.did.sdk.crypto.keys.TestKeys
 import com.microsoft.did.sdk.crypto.keys.ellipticCurve.EllipticCurvePairwiseKey
+import com.microsoft.did.sdk.crypto.keys.ellipticCurve.EllipticCurvePublicKey
 import com.microsoft.did.sdk.crypto.models.Sha
-import com.microsoft.did.sdk.crypto.models.webCryptoApi.CryptoKeyPair
-import com.microsoft.did.sdk.crypto.models.webCryptoApi.KeyUsage
-import com.microsoft.did.sdk.crypto.models.webCryptoApi.W3cCryptoApiConstants
+import com.microsoft.did.sdk.crypto.models.webCryptoApi.*
 import com.microsoft.did.sdk.crypto.models.webCryptoApi.algorithms.EcKeyGenParams
+import com.microsoft.did.sdk.crypto.protocols.jose.jws.JwsToken
+import com.microsoft.did.sdk.datasource.db.SdkDatabase
+import com.microsoft.did.sdk.datasource.db.dao.IdentifierDao
+import com.microsoft.did.sdk.datasource.network.apis.ApiProvider
 import com.microsoft.did.sdk.di.defaultTestSerializer
 import com.microsoft.did.sdk.identifier.IdentifierCreator
 import com.microsoft.did.sdk.identifier.SidetreePayloadProcessor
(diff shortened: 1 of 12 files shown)


sdk/src/androidTest/java/com/microsoft/did/sdk/crypto/protectors/TokenSignerPerformanceTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        when (result) {
+            is Result.Success -> {
+                // use result.payload (Identifier)
+                val signedPayload = signer.signWithIdentifier(payload, result.payload)
+                println("PerfTest->(${getTestName()}) - End sign: ${timer(startTime)}")
+                println("PerfTest->(${getTestName()}) - Signed token: ${signedPayload}")
+            }
+        }
+
+        println("PerfTest->(${getTestName()}) - Start verify")
```
nithyaganeshng, line 97: TokenSigner class does only signing I believe. Can we remove print statements for verification here?

sdk/src/androidTest/java/com/microsoft/did/sdk/crypto/protectors/TokenSignerPerformanceTest.kt
```diff
@@ -0,0 +1,103 @@
...
+                    "KeyReference" to keyReference
+                )
+            ), true, listOf(KeyUsage.Sign)
+        )
+
+        cryptoOperations = CryptoOperations(ellipticCurveSubtleCrypto, keyStore, EllipticCurvePairwiseKey())
+        val sidetreePayloadProcessor = SidetreePayloadProcessor(defaultTestSerializer)
+        identifierCreator = IdentifierCreator(cryptoOperations, sidetreePayloadProcessor)
+
+        signer = TokenSigner(cryptoOperations, serializer)
```
gproanomsft, line 64: I don't know if authenticator uses ahead of time or just in time compilation, but generally running once makes for  an inaccurate perf assessment, in the case of JIT - you spend a lot of time not executing your code

absent a dedicated benchmarking framework, in the past I would benchmark by running the code once to ensure any JIT occurs, and then running in a 10k loop

sdk/src/main/java/com/microsoft/did/sdk/crypto/keyStore/AndroidKeyStore.kt
```diff
@@ -57,53 +65,88 @@ class AndroidKeyStore @Inject constructor(private val context: Context, private
...
-        val nativeKeys = listNativeKeys()
-        var key = nativeKeys[keyReference]
-        if (key != null) {
-            return KeyContainer(
-                kty = key.kty,
-                keys = key.kids.map {
-                    AndroidKeyConverter.androidPrivateKeyToPrivateKey(it, keyStore)
-                }
-            )
+        var privateKey = nativePrivateKeyCache.get(keyReference)
```
dangodb, line 68: Are we using both the native and software keystores at this point? if not, let's get rid of one of them. We can reintroduce it later when we need it

sdk/src/main/java/com/microsoft/did/sdk/crypto/keyStore/AndroidKeyStore.kt
```diff
@@ -35,6 +35,14 @@ class AndroidKeyStore @Inject constructor(private val context: Context, private
 
     companion object {
         const val provider = "AndroidKeyStore"
+        // caches
+        private val nativePrivateKeyCache = HashMap<String, KeyContainer<PrivateKey>>()
```
dangodb, line 39: How long does the AndroidKeyStore lives for in memory? Are we concerned about keeping the secrets in clear in memory for that long? Would it be better to only warmup the required secrets and dispose of them when the signature is done?

sdk/src/main/java/com/microsoft/did/sdk/crypto/keyStore/AndroidKeyStore.kt
```diff
@@ -231,10 +277,15 @@ class AndroidKeyStore @Inject constructor(private val context: Context, private
                 }
             }
         }
+        nativeList = output
         return output
     }
 
     private fun listSecureData(): Map<String, KeyStoreListItem> {
+        if (secureDataList != null) {
```
dangodb, line 285: Does this has a big perf impact? IIRC sharedPrefs does a load from disk the 1st time and then use the inmemory cache for subsequent fetch. Does EncryptedSharedPreferences has a similar concept? If so, this guy may not be needed. I'm a bit concerned about the number of caches that we have in this class.

dangodb (changes requested):

AndiMiko:
Obsolete due to crypto rework

PR #137, opened 2020-12-28
Niganesh/adding correlation vector

nithyaganeshng:
Adding correlation vector between network calls to help correlate different interactions in a flow.

**Validation:**

- Verified that all the tests pass successfully.
- Manually tested various scenarios to verify if correlation id was created and deleted properly
- Manually verified telemetry data for correlation id

**Type of change:**
- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:


sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
```diff
@@ -41,6 +41,7 @@ abstract class BaseNetworkOperation<S, T> {
         // TODO("how do we want to handle null bodies")
         // TODO("how to not suppress this warning")
         @Suppress("UNCHECKED_CAST")
+        val correlationVector = response.headers()["ms-cv"] ?: "?"
```
symorton, line 44: why should it be equal to "?"  if response.headers()["ms-cv"] is null?
Also should "ms-cv" be in a constants file?

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
```diff
@@ -0,0 +1,37 @@
...
+        val requestWithCorrelationVector = requestWithCorrelationVectorBuilder.build()
+        return chain.proceed(requestWithCorrelationVector)
+    }
+
+    private fun writeCorrelationVector(applicationContext: Context, correlationId: String) {
+        PreferenceManager.getDefaultSharedPreferences(applicationContext).edit().putString("correlation_vector", correlationId).apply()
+    }
+
+    private fun readCorrelationVector(applicationContext: Context): String? {
+        return PreferenceManager.getDefaultSharedPreferences(applicationContext).getString("correlation_vector", null)
```
symorton, line 35: nit: should "correlation_vector" be in a constants file maybe?

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
```diff
@@ -0,0 +1,37 @@
...
+import okhttp3.Interceptor
+import okhttp3.Interceptor.Chain
+import okhttp3.Response
+import java.io.IOException
+import javax.inject.Inject
+
+class CorrelationVectorInterceptor @Inject constructor(private val context: Context) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Chain): Response {
```
symorton, line 17: so this Interceptor is getting the correlation vector, and incrementing it if not null? And if it is null, providing a new one?

dangodb: I don't see where the initial vector is created, can you point me to the file/line?

dangodb (changes requested):

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
nithyaganeshng: This is existing code. But you are right, it makes sense to move names of headers to constants file. I will make the change.

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
nithyaganeshng: Correlation id is created in wallet code while initiating the request. So code changes for it is in that PR. SDK only reads and increments it if it exists. If wallet doesn't provide one, we are not creating one in SDK now. Do we want to have control of correlation creation in both wallet and SDK?

sdk/build.gradle
```diff
@@ -111,6 +115,7 @@ dependencies {
     implementation "com.madgag.spongycastle:prov:1.58.0.0"
     //Used for Json Canonicalization performed for ION
     implementation "io.github.erdtman:java-json-canonicalization:1.1"
+    implementation 'com.github.microsoft:CorrelationVector-Java:master-SNAPSHOT'
```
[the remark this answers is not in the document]
nithyaganeshng: I will change the quotes. Unfortunately the library is not published with versions. I tried it and it fails trying to find it.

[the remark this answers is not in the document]
nithyaganeshng: Never mind, I was able to use v1.0. I was using a different artifact and group id from build but was able to find the correct one now.

sdk/build.gradle
```diff
@@ -12,6 +12,10 @@ jacoco {
     toolVersion = "0.8.5"
 }
 
+repositories{
+    maven { url 'https://jitpack.io' }
```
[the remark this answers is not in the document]
nithyaganeshng: Yes that's the reason.

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
nithyaganeshng: We currently follow a pattern to populate any missing information with ? in Android be it claim details or header information. But I don't think we have any attachment to it.

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
AndiMiko: This interceptor seems to intercept any network call, while creating a new CV is the responsibility of the wallet. As I do not see any "end" or "reset" option for the CV, wouldn't this mean that after making a few calls for the "CardFlow" part all subsequent calls that are unrelated to the CardFlow (e.g. revoking something within Card Details after ending the CardFlow) would still be incrementing the same CV and thus appear correlated?

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.microsoft.did.sdk.util.Constants
+import javax.inject.Inject
+
+class CorrelationVectorService @Inject constructor(private val context: Context) {
+
+    fun createAndSave(): String {
+        val correlationId = CorrelationVector(CorrelationVectorVersion.V2).value
+        saveCorrelationVector(context, correlationId)
+        return correlationId
+    }
```
AndiMiko, line 21: nit, maybe call it something like StartNew to make it explicit that this is supposed to be called every time a new correlation is getting started?

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
AndiMiko: For "independent" requests like a revoke, does it make sense to have a unique CV or not?

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    fun incrementAndSave(): String {
+        val correlationVectorString = getCorrelationVector(context)
+        if (correlationVectorString != null) {
+            val correlationVectorIncremented = CorrelationVector.parse(correlationVectorString).increment()
+            saveCorrelationVector(context, correlationVectorIncremented)
+            return correlationVectorIncremented
+        }
+        return ""
```
AndiMiko, line 30: Is this a valid case? Not sure what it means if no CV is available but you are still trying to increment it. Is this an exception? Or is this ok?

Would need to understand all scenarios for that

AndiMiko (changes requested):

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
```diff
@@ -0,0 +1,27 @@
...
+import okhttp3.Interceptor
+import okhttp3.Interceptor.Chain
+import okhttp3.Response
+import java.io.IOException
+import javax.inject.Inject
+
+class CorrelationVectorInterceptor @Inject constructor(private val correlationVectorService: CorrelationVectorService) : Interceptor {
+
+    @Throws(IOException::class)
+    override fun intercept(chain: Chain): Response {
```
dangodb, line 17: Yep, with the current code, the responsibility of starting a new flow (generating a new CV) belongs to the wallet. In your example, the wallet has the responsibility of starting a new flow when a recvocation happens.

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
dangodb: +1, maybe StartNewFlow()?

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
```diff
@@ -0,0 +1,41 @@
...
+
+class CorrelationVectorService @Inject constructor(private val context: Context) {
+
+    fun createAndSave(): String {
+        val correlationId = CorrelationVector(CorrelationVectorVersion.V2).value
+        saveCorrelationVector(context, correlationId)
+        return correlationId
+    }
+
+    fun incrementAndSave(): String {
```
dangodb, line 23: Only createAndSave needs to be public, no?

dangodb (approved):

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
nithyaganeshng: If correlation vector is null or empty, we do not increment it. writing the correlation vector is controlled by SDK only and I have added a check to not allow empty values to be written.

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
nithyaganeshng: It makes sense to have separate CVs for those kind of calls. Right now we are not clearing the CV instead it is overwritten when a new request is created. We could do the same for these calls as well. Creating a new CV and overwriting the existing one can be done in wallet wherever the request is initiated for any type of calls.

sdk/src/main/java/com/microsoft/did/sdk/di/CorrelationVectorInterceptor.kt
nithyaganeshng: 👍

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
nithyaganeshng: yes, I made this internal since it is needed only within SDK.

sdk/src/main/java/com/microsoft/did/sdk/CorrelationVectorService.kt
nithyaganeshng: created a default cv while initializing SDK so there is always a cv in there.

AndiMiko (approved):

PR #139, opened 2021-01-26
Warn instead of fail if well known config document is not uploaded.

nithyaganeshng:
If domain is bound to DID but config document is not uploaded, consider it a warning case instead of failure. 

**Solution:**
Changed the result returned as warning instead of failure. Also completely bypass linked domain verification if feature flag is turned off.
As part of this, made the following changes.
- Created a class for feature flags
- Replaced injecting the context with only shared preference as it is sufficient

**Validation:**
Manually verified success and failure scenarios.
Verified if all the tests successfully passed.

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
@@ -9,6 +9,7 @@ import com.microsoft.did.sdk.credential.models.VerifiableCredential
 import com.microsoft.did.sdk.credential.service.IssuanceRequest
 import com.microsoft.did.sdk.credential.service.IssuanceResponse
 import com.microsoft.did.sdk.credential.service.RequestedVcMap
+import com.microsoft.did.sdk.credential.service.models.linkedDomains.LinkedDomainDisabled
 import com.microsoft.did.sdk.credential.service.protectors.IssuanceResponseFormatter
 import com.microsoft.did.sdk.credential.service.validators.JwtValidator
 import com.microsoft.did.sdk.datasource.network.apis.ApiProvider
@@ -38,11 +39,12 @@ class IssuanceService @Inject constructor(
      *
      * @param contractUrl url that the contract is fetched from
      */
-    suspend fun getRequest(contractUrl: String, isLinkedDomainsEnabled: Boolean): Result<IssuanceRequest> {
+    suspend fun getRequest(contractUrl: String): Result<IssuanceRequest> {
         return runResultTry {
             val contract = fetchContract(contractUrl).abortOnError()
+            val isLinkedDomainsEnabled = VerifiableCredentialSdk.FeatureFlag.linkedDomains
             val linkedDomainResult =
-                if (isLinkedDomainsEnabled) linkedDomainsService.fetchAndVerifyLinkedDomains(contract.input.issuer).abortOnError() else null
+                if (isLinkedDomainsEnabled) linkedDomainsService.fetchAndVerifyLinkedDomains(contract.input.issuer).abortOnError() else LinkedDomainDisabled()
             val request = IssuanceRequest(contract, contractUrl, linkedDomainResult)
             Result.Success(request)
         }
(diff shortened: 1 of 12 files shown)


sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
```diff
@@ -38,10 +38,11 @@ class IssuanceService @Inject constructor(
      *
      * @param contractUrl url that the contract is fetched from
      */
-    suspend fun getRequest(contractUrl: String): Result<IssuanceRequest> {
+    suspend fun getRequest(contractUrl: String, isLinkedDomainsEnabled: Boolean): Result<IssuanceRequest> {
```
AndiMiko, line 41: This is a global flag, not really called/changed on a per call basis right?

If we add these kind of feature flags to our public API it would change often. Also I think it already looks not great.

How about making this a global setting instead when initializing the SDK?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
```diff
@@ -13,13 +13,13 @@ import com.microsoft.did.sdk.credential.service.models.presentationexchange.Pres
...
 @Serializable
-sealed class Request(val entityName: String, val entityIdentifier: String, val entityLinkedDomainResult: LinkedDomainResult)
+sealed class Request(val entityName: String, val entityIdentifier: String, val entityLinkedDomainResult: LinkedDomainResult? = null)
 
 @Serializable
 class IssuanceRequest(
     val contract: VerifiableCredentialContract,
     val contractUrl: String,
-    val linkedDomainResult: LinkedDomainResult
+    val linkedDomainResult: LinkedDomainResult? = null
```
AndiMiko, line 22: instead of making it nullable, isn't this what `LinkedDomainMissing` is for? If this has indeed a different meaning and you don't want to further overload it I would rather create another class then.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
```diff
@@ -13,13 +13,13 @@ import com.microsoft.did.sdk.credential.service.models.presentationexchange.Pres
 import kotlinx.serialization.Serializable
 
 @Serializable
-sealed class Request(val entityName: String, val entityIdentifier: String, val entityLinkedDomainResult: LinkedDomainResult)
+sealed class Request(val entityName: String, val entityIdentifier: String, val entityLinkedDomainResult: LinkedDomainResult? = null)
```
AndiMiko, line 16: nit, default value shouldn't be needed here as it's only called by it's two child classes bellow (they always pass their default value)

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
AndiMiko: actually the default value isn't needed here as well or is there any caller that doesn't specify it? I would keep it explicit probably

AndiMiko (changes requested):

sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
nithyaganeshng: It couldn't set while initializing as it could vary depending on the feature flag now. It may be turned on and off.

nithyaganeshng: Created a class for Feature flag and injected it now.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
nithyaganeshng: Added a separate class for Disabled type.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/Request.kt
nithyaganeshng: it is not nullable now, that fixes it.

AndiMiko (approved):

PR #143, opened 2021-02-05
Improve error message and exceptions.

nithyaganeshng:
**Problem:**
Client exception name was incorrect and consolidated many scenarios. Was sending out huge error message.

**Solution:**
Fixed confusing exception names and added explicit exceptions for client error codes. Added better error messages and split error message sent into smaller chunks to reduce its size.

**Validation:**

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [x] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
@@ -56,7 +56,8 @@ abstract class BaseNetworkOperation<S, T> {
                 UnauthorizedException(
                     requestId,
                     correlationVector,
-                    defaultErrorMessage(response.code(), requestId, correlationVector, response.errorBody()?.string() ?: ""),
+                    response.code().toString(),
+                    response.errorBody()?.string() ?: "",
                     false
                 )
             )
@@ -64,7 +65,8 @@ abstract class BaseNetworkOperation<S, T> {
                 ServiceErrorException(
                     requestId,
                     correlationVector,
-                    defaultErrorMessage(response.code(), requestId, correlationVector, response.errorBody()?.string() ?: ""),
+                    response.code().toString(),
+                    response.errorBody()?.string() ?: "",
                     false
                 )
             )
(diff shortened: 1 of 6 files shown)


sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
```diff
@@ -56,40 +58,53 @@ abstract class BaseNetworkOperation<S, T> {
...
+            403 -> {
+                Result.Failure(
+                    ForbiddenException(
+                        requestId,
+                        correlationVector,
+                        response.code().toString(),
+                        response.errorBody()?.string() ?: "",
+                        false
+                    )
+                )
```
AndiMiko, line 84: Not sure if we really should create new exceptions for every status code. I mean in some ways it's ok, but then if you think we would add one for each status code it blows up. I think at that points it's easier to just pass the response code and it's enough.

This is ok for this PR, we will be working on other status codes anyhow I believe

AndiMiko (approved):

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
nithyaganeshng: I don't know if we need 402. But for 403 and 404, I wanted to make 404 retryable since it is temporary and happens a lot during resolution. And I now see that I haven't made it retryable, I will set it to true.

nithyaganeshng: Also for these common ones, I wanted to process the error response rather than using the whole error response but there are not complete on services yet. So I will revisit this code to process those when ready.

AndiMiko: My point is that you can make them retryable by supplying retryable to a common exception type. You can argue both ways

PR #144, opened 2021-02-08
Rework Crypto/KeyStore/JWS/Pairwise/Sidetree

AndiMiko:
The crypto layer has grown from a proof-of-concept stage and has been largely unmaintained since then. This caused a variety of issues. Most notably performance, code quality and maintainability, test coverage and lack of confidence in security fundamentals.

This PR reworks/rewrites large parts of these areas:
- **Keystore.** Keys are now stored in EncryptedSharedPreferences which are encrypted by hardware module backed keys. We have explored using hardware backed keys directly but decided they are not suitable for our requirements.
- **Crypto.** Removed WebCrypto and replaced it with Javas default security framework. `CryptoOperations` is now purely a type safety and convenience provider for Java security.
- **Protocol**. `JwsTokens` are now driven by Nimbus.
- **Pairwise**. Pairwise functionality is rewritten and part of Java Security.
- **SideTree**. Much of SideTree/ION has been reworked.
- Replaced standard functions like Base64 encoding and type conversions (ByteArrays etc.) by default language implementations.
- Performance should be significantly increased. (50x+)


sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved
+
+package com.microsoft.did.sdk.util
+
+import com.microsoft.did.sdk.util.log.SdkLog
+
+inline fun <R> logTime(name: String, block: () -> R): R {
```
nithyaganeshng, line 7: That's a sweet function. 👌

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/linkedDomains/LinkedDomainResult.kt
```diff
@@ -14,4 +14,4 @@ class LinkedDomainVerified(val domainUrl: String) : LinkedDomainResult()
 class LinkedDomainUnVerified(val domainUrl: String) : LinkedDomainResult()
 
 @Serializable
-class LinkedDomainMissing : LinkedDomainResult()
\ No newline at end of file
+object LinkedDomainMissing : LinkedDomainResult()
```
nithyaganeshng, line 17: I remember I had an issue or had to do something differently on wallet side while handling it if I declared this alone as an object. Just make sure this case works fine.

sdk/src/main/java/com/microsoft/did/sdk/identifier/IdentifierCreator.kt
```diff
@@ -4,189 +4,140 @@
...
 
-    private fun computeLongFormIdentifier(registrationPayload: RegistrationPayload): String {
-        val registrationPayloadCanonicalized =
-            canonicalizeAsByteArray(serializer.encodeToString(RegistrationPayload.serializer(), registrationPayload))
-        val registrationPayloadCanonicalizedEncoded = Base64Url.encode(registrationPayloadCanonicalized)
-        val identifierShortForm = computeUniqueSuffix(registrationPayload)
-        return "$identifierShortForm${Constants.COLON}$registrationPayloadCanonicalizedEncoded"
+    internal fun generatePersonaSeed(personaDid: String): ByteArray {
+        val masterSeed = keyStore.getKey(Constants.MASTER_IDENTIFIER_NAME).toOctetSequenceKey().toByteArray()
+        return CryptoOperations.computeMac(personaDid.toByteArray(), SecretKeySpec(masterSeed, "AES"), MacAlgorithm.HmacSha512)
```
nithyaganeshng, line 85: nit: Can we move the "AES" algorithm name to a constants file?

sdk/src/main/java/com/microsoft/did/sdk/crypto/spi/EcPairwiseKeyFactorySpi.kt
```diff
@@ -0,0 +1,87 @@
...
+        val qNormalized = q.normalize()
+        return CryptoOperations.generateKey(
+            PublicKeyFactoryAlgorithm.Secp256k1(
+                qNormalized.affineXCoord.toBigInteger(),
+                qNormalized.affineYCoord.toBigInteger()
+            )
+        )
+    }
+
+    override fun engineGeneratePrivate(keySpec: KeySpec?): PrivateKey {
```
nithyaganeshng, line 38: Could you please add some comments to this method and above to explain what it does?

sdk/src/main/java/com/microsoft/did/sdk/crypto/spi/EcPairwiseKeyFactorySpi.kt
```diff
@@ -0,0 +1,87 @@
...
+    override fun engineGeneratePublic(keySpec: KeySpec?): PublicKey {
+        val ecKeySpec = keySpec as? EcPairwisePublicKeySpec
+            ?: throw InvalidKeySpecException("Keyspec has to be of type ${EcPairwisePublicKeySpec::class.qualifiedName}")
+
+        val ecSpec: ECParameterSpec = ECNamedCurveTable.getParameterSpec(Constants.SECP256K1_CURVE_NAME_EC)
+        val q: ECPoint = ecSpec.g.multiply(ecKeySpec.privateKey.s)
+        val qNormalized = q.normalize()
+        return CryptoOperations.generateKey(
+            PublicKeyFactoryAlgorithm.Secp256k1(
+                qNormalized.affineXCoord.toBigInteger(),
```
nithyaganeshng, line 32: Can you verify if this consistently passes signature check?

sdk/src/main/java/com/microsoft/did/sdk/identifier/SideTreeHelper.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.microsoft.did.sdk.crypto.CryptoOperations
+import com.microsoft.did.sdk.crypto.DigestAlgorithm
+import com.microsoft.did.sdk.util.Constants
+import com.nimbusds.jose.jwk.JWK
+import org.erdtman.jcs.JsonCanonicalizer
+import javax.inject.Inject
+
+class SideTreeHelper @Inject constructor() {
+
+    fun canonicalizeAndMultiHash(json: String): String {
```
nithyaganeshng, line 15: change the name to reflect encoding is done as well as it is one of the major steps for it to match on ION.

sdk/src/main/java/com/microsoft/did/sdk/identifier/SideTreeHelper.kt
```diff
@@ -0,0 +1,46 @@
...
+        val jsonCanonicalizer = JsonCanonicalizer(json)
+        val hashed = CryptoOperations.digest(jsonCanonicalizer.encodedUTF8, DigestAlgorithm.Sha256)
+        val hashedInfo = prependMultiHashInfo(hashed)
```
nithyaganeshng, line 18: nit: can you change names for both hashedInfo and doubleHashedInfo in the method below to include multihash.

sdk/build.gradle
```diff
@@ -108,12 +108,13 @@ dependencies {
...
-    implementation "androidx.preference:preference:$androidx_preference_version"
+    implementation "androidx.preference:preference-ktx:$androidx_preference_version"
     implementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
 
     api 'com.google.guava:guava:28.1-android'
-    // TODO: this seems to be required as api for org.bitcoin.jar? Can we avoid this?
-    implementation "com.madgag.spongycastle:prov:1.58.0.0"
+    implementation "org.bouncycastle:bcprov-jdk15to18:$org_bouncycastle"
+    // JWE/JWS
+    implementation "com.nimbusds:nimbus-jose-jwt:$nimbus_jose_jwt_version"
```
nithyaganeshng, line 117: Can you make sure it is added to CG?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/ExchangeResponseFormatter.kt
```diff
@@ -34,12 +33,12 @@ class ExchangeResponseFormatter @Inject constructor(
         responseId: String
     ): String {
         val requester = exchangeRequest.requester
-        val key = cryptoOperations.keyStore.getPublicKey(requester.signatureKeyReference).getKey()
+        val keyJwk = keyStore.getKey(requester.signatureKeyReference)
```
nithyaganeshng, line 36: can we have a method to get the public key which we can use everywhere else where signing is not performed so that we don't pass around private key? In other places we need only the public key.

sdk/src/test/java/android/util/Base64.kt
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft Corporation. All rights reserved
+@file:JvmName("Base64")
+
+package android.util
```
nithyaganeshng, line 4: I don't see this class being used anywhere. Do we need this for tests to run?

sdk/src/test/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/JwsTokenTest.kt
```diff
@@ -1,84 +1,51 @@
...
-
-    private val ellipticCurvePairwiseKey = EllipticCurvePairwiseKey()
+    private val keyRef = Base64.getEncoder().encodeToString(Random.nextBytes(8))
+    private val key: ECKey
+    private val payload: String = "{\"iss\":\"joe\",\n" +
+        " \"exp\":1300819380,\n" +
+        " \"http://example.com/is_root\":true}"
 
     init {
         /* This is the payload used for all the operations below */
```
nithyaganeshng, line 19: nit: outdated comment.

sdk/src/test/java/com/microsoft/did/sdk/identifier/SidetreePayloadProcessorTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    @Test
+    fun `update commitment is properly calculated from updateJwk`() {
+        val expectedCommitment = "EiCrZyAhnMP-VM-m-C1-6VrCmwIka6fR0yDXQ1WIbjzbzg"
+        val actualRegistrationPayload = sidetreePayloadProcessor.generateCreatePayload(signingJwk, recoverJwk, updateJwk)
+        assertThat(actualRegistrationPayload.patchData.nextUpdateCommitmentHash).isEqualTo(expectedCommitment)
+    }
+
+    @Test
+    fun `recovery commitment is properly calculated from recoveryJwk`() {
+        val expectedCommitment = "EiDSwd8T1Uzd0gRKQ0MuBUaxdKlM_1GIO-NNEXaiW-4AqQ"
```
nithyaganeshng, line 37: nit: I believe it is the commitment hash being compared.

sdk/src/test/java/com/microsoft/did/sdk/identifier/SideTreeHelperTest.kt
```diff
@@ -0,0 +1,46 @@
...
+
+package com.microsoft.did.sdk.identifier
+
+
+import com.nimbusds.jose.jwk.JWK
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class SideTreeHelperTest {
+
```
nithyaganeshng, line 11: 👌 I really like these tests 😊

sdk/src/test/java/com/microsoft/did/sdk/identifier/SidetreePayloadProcessorTest.kt
```diff
@@ -0,0 +1,49 @@
...
+class SidetreePayloadProcessorTest {
+
+    private val sidetreePayloadProcessor = SidetreePayloadProcessor(SideTreeHelper(), defaultTestSerializer)
+
+    private val signingJwk = JWK.parse("{\"kty\":\"EC\",\"crv\":\"secp256k1\",\"kid\":\"signKey\",\"x\":\"dqEeSHC5KhsMSW_Zh8kBzQXB9HLgZqzBtmkAh-tAw4U\",\"y\":\"Yo_a4_sB2METsA9YRD6II_PjbHiWg4gwqQJiOxx4Suk\"}")
+    private val recoverJwk = JWK.parse("{\"kty\":\"EC\",\"crv\":\"secp256k1\",\"kid\":\"recoverKey\",\"x\":\"weF3fxy4XqvMLpqYBROUMNf9q7MWpazuw4J5dEvsOO0\",\"y\":\"vxRL166b7Nv66bLI6EocgdNzGtiD_k-vEluRrZcEf-k\"}")
+    private val updateJwk = JWK.parse("{\"kty\":\"EC\",\"crv\":\"secp256k1\",\"kid\":\"updateKey\",\"x\":\"iK9EveBFpO_KltBpKJfvq2KA2da-_VmYOwhJCAk6pRM\",\"y\":\"45wg6vkIT4JXcwzEJBjEWsJWxWv9cLXhdOAA4CACb8k\"}")
+
+    @Test
+    fun `signing key is assigned to patchData correctly`() {
```
nithyaganeshng, line 20: Could we also similar tests for recovery and update keys? may be I am getting too excited here since we have a good coverage 😁

sdk/src/main/java/com/microsoft/did/sdk/crypto/keyStore/EncryptedKeyStore.kt
```diff
@@ -0,0 +1,42 @@
...
+        return EncryptedSharedPreferences.create(
+            FILE_NAME,
+            masterKeyAlias,
+            context,
+            EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+            EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
+        )
+    }
+
+    fun storeKey(key: JWK, keyId: String) {
```
nithyaganeshng, line 33: just curious, it is like key/value pair right. we generally do key as 1st argument, don't we or is it just me.

sdk/src/test/java/com/microsoft/did/sdk/identifier/IdentifierCreatorTest.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    private fun createKeyPair(s: String, x: String, y: String): KeyPair {
+        val privateKey = CryptoOperations.generateKey<ECPrivateKey>(PrivateKeyFactoryAlgorithm.Secp256k1(BigInteger(s)))
+        val publicKey = CryptoOperations.generateKey<ECPublicKey>(PublicKeyFactoryAlgorithm.Secp256k1(BigInteger(x), BigInteger(y)))
+        return KeyPair(publicKey, privateKey)
+    }
+
+    @Test
+    fun `pairwise Identifier gets created properly`() {
```
nithyaganeshng, line 97: Not needed right away but we could add couple more tests to verify if it is same long form and same pairwiseIdName everytime.

nithyaganeshng (changes requested):
Looks like my latest merges caused a conflict. I have left some minor comments but it looks good to me.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/models/linkedDomains/LinkedDomainResult.kt
AndiMiko: now that you mentioned it I recall we talked about it briefly? I kinda just stumbled over it and changed it. Will change back if we find issues :)

sdk/src/main/java/com/microsoft/did/sdk/identifier/IdentifierCreator.kt
AndiMiko: done

sdk/src/main/java/com/microsoft/did/sdk/crypto/spi/EcPairwiseKeyFactorySpi.kt
AndiMiko: I added a class comment. I believe the comments you are looking for are already on the overriden methods. Not sure if it's a good practice to copy them over here.

sdk/src/main/java/com/microsoft/did/sdk/crypto/spi/EcPairwiseKeyFactorySpi.kt
AndiMiko: Isn't this tested in `PairwiseKeyTest.generateSameKeysInFile()`?

sdk/src/main/java/com/microsoft/did/sdk/identifier/SideTreeHelper.kt
AndiMiko: done

sdk/src/main/java/com/microsoft/did/sdk/identifier/SideTreeHelper.kt
AndiMiko: where? 😅

sdk/build.gradle
AndiMiko: good call. Are you referring to Authenticators or do we have our own by now?

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/ExchangeResponseFormatter.kt
AndiMiko: well I see your point, but not sure whether there is an elegant way to solve this. Because if you add for example `keyStore.getPublicKey()`
a) someone could still call `keyStore.getKey()`
b) it would be the same outcome as `keyStore.getKey().toPublicJWK()` (which is fine to be used). But if this would be coded inside the keyStore method then it would assume that the key is a private key. What if a publicKey was stored there?

Not sure if we get value out of that. But feel free to make a suggestion if you see a good way of adding that

sdk/src/test/java/android/util/Base64.kt
AndiMiko: This is needed to mock an android package class within tests. Really weird there is no better solution ... however, thanks for noting! It was really unclear. I added a suppression and class comment to explain what this is.

sdk/src/test/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/JwsTokenTest.kt
AndiMiko: and that's why comments are bad :P thanks. (btw some of the code is written by Logan but I'll make the changes)

sdk/src/test/java/com/microsoft/did/sdk/identifier/SidetreePayloadProcessorTest.kt
AndiMiko: IIRC a commitment is always a hash. (commentmentHash and revealValue)

sdk/src/test/java/com/microsoft/did/sdk/identifier/SidetreePayloadProcessorTest.kt
AndiMiko: isn't that what the two next tests do? Not sure if I understand what you mean

sdk/src/main/java/com/microsoft/did/sdk/crypto/keyStore/EncryptedKeyStore.kt
AndiMiko: you are totally right!

sdk/src/test/java/com/microsoft/did/sdk/identifier/IdentifierCreatorTest.kt
AndiMiko: Is there reason to believe that the creation might not be deterministic? I hardly see how it would not be. More so, wouldn't this imply this test would periodically fail if it's not deterministic?

Maybe I'm missing the point of such a test

sdk/src/main/java/com/microsoft/did/sdk/crypto/spi/EcPairwiseKeyFactorySpi.kt
nithyaganeshng: ah this is pairwise, yeah that test passing should verify this. We were doing similar manipulation of getting a public key from random bytes (private key). That was causing the issues earlier. I see now we generate a key pair for secp256k1 keys instead of doing the manipulation so no issue now. got confused there.

sdk/src/main/java/com/microsoft/did/sdk/crypto/spi/EcPairwiseKeyFactorySpi.kt
nithyaganeshng: Overriden methods would have comments on what this method is supposed to do. But our implementation might have specifics like create a mac, reduce the size and so on as in this case and we don't have the control to name the method to reflect that. So was wondering if comments could help. But in general I agree to not having too many comments as they easily get outdated.

sdk/src/main/java/com/microsoft/did/sdk/identifier/SideTreeHelper.kt
nithyaganeshng: line 18 and 36.

sdk/build.gradle
nithyaganeshng: We have our own. I will send you the link to it.

sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/ExchangeResponseFormatter.kt
nithyaganeshng: I am thinking if we can make getKey() a private method? if it should only be used for signing and make it part of that class so that anyone who wants to sign can only send the payload, get it signed and receive a jwt but not access the key. But any other class can get the public key.

sdk/src/test/java/android/util/Base64.kt
nithyaganeshng: ah okay. I was doing a find usage on this class in IDE and didn't find any so got confused.

sdk/src/test/java/com/microsoft/did/sdk/identifier/SidetreePayloadProcessorTest.kt
nithyaganeshng: 🤔 I have never thought about it but always used commitmenthash in all the conversations. It totally makes sense now.

sdk/src/test/java/com/microsoft/did/sdk/identifier/SidetreePayloadProcessorTest.kt
nithyaganeshng: yeah right my bad, we only have commitments in the payload for those keys, I forgot that.

sdk/src/test/java/com/microsoft/did/sdk/identifier/IdentifierCreatorTest.kt
nithyaganeshng: Actually this test includes testing both the methods I had suggested. I was assuming this test to just verify if pairwise id was successfully created. It actually verifies the hashed pairwise name and long form created.

sdk/src/androidTest/java/com/microsoft/did/sdk/crypto/keyStore/EncryptedKeyStoreInstrumentedTest.kt
```diff
@@ -0,0 +1,44 @@
...
+    fun saveAndGet() {
+        keyStore.storeKey("key1", signingJwk)
+        val actualKey = keyStore.getKey("key1")
+        assertThat(signingJwk).isEqualToComparingFieldByFieldRecursively(actualKey)
+    }
+
+    @Test
+    fun overwriteKey() {
+        keyStore.storeKey("key2", signingJwk)
+        keyStore.storeKey("key2", updateJwk)
```
beejones, line 34: you should use a key container keys: [] so you store a new version of the key (rolling) instead of overwrite.

sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
```diff
@@ -66,17 +69,19 @@ class IssuanceService @Inject constructor(
         enablePairwise: Boolean = true
     ): Result<VerifiableCredential> {
         return runResultTry {
-            val masterIdentifier = identifierManager.getMasterIdentifier().abortOnError()
-            val verifiableCredential = if (enablePairwise) {
```
beejones, line 70: Did we keep the notion of personas? Are pairwise enabled/disabled per persona?

sdk/src/main/java/com/microsoft/did/sdk/crypto/Algorithms.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.nimbusds.jose.jwk.Curve
+import java.math.BigInteger
+import java.security.spec.AlgorithmParameterSpec
+import java.security.spec.ECPoint
+import java.security.spec.ECPrivateKeySpec
+import java.security.spec.ECPublicKeySpec
+import java.security.spec.KeySpec
+
+abstract class SigningAlgorithm(val name: String, val provider: String?, val spec: AlgorithmParameterSpec? = null) {
+    object Ecdsa256 : SigningAlgorithm("SHA256withECDSA", null) // EXAMPLE
```
beejones, line 16: ecdsa256 is not a good algorithm identifier because it does not specify the curve. Better to use jwa values, see https://tools.ietf.org/html/rfc7518#page-9
ES256K

sdk/src/main/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/serialization/JwkSerializer.kt
```diff
@@ -0,0 +1,90 @@
...
+        val octJwk = value as? OctetSequenceKey
+        val okpJwk = value as? OctetKeyPair
+        encoder.encodeSerializableValue(
+            JwkSurrogate.serializer(), JwkSurrogate(
+                kty = value.keyType.value,
+                kid = value.keyID,
+                use = value.keyUse?.value,
+                key_ops = value.keyOperations?.toList()?.map { it.identifier() },
+                alg = value.algorithm?.name,
+                crv = ecJwk?.curve?.stdName ?: okpJwk?.curve?.stdName,
```
beejones, line 31: the mixing of key types is error prone. kty defines the key type (RSA, EC, OKP). The properties should be set accordingly.

sdk/src/main/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/serialization/JwkSerializer.kt
```diff
@@ -0,0 +1,90 @@
...
+object JwkSerializer : KSerializer<JWK> {
+    override val descriptor: SerialDescriptor = JwkSurrogate.serializer().descriptor
+    override fun serialize(encoder: Encoder, value: JWK) {
+        val ecJwk = value as? ECKey
+        val rsaJwk = value as? RSAKey
+        val octJwk = value as? OctetSequenceKey
+        val okpJwk = value as? OctetKeyPair
+        encoder.encodeSerializableValue(
+            JwkSurrogate.serializer(), JwkSurrogate(
+                kty = value.keyType.value,
```
beejones, line 26: same comment here. kty should define the serialization.

sdk/src/main/java/com/microsoft/did/sdk/identifier/IdentifierCreator.kt
```diff
@@ -4,189 +4,139 @@
...
-    private fun transformIdentifierDocumentToIdentifier(
-        identifierLongForm: String,
-        alias: String,
-        signatureKeyReference: String,
-        recoveryKeyReference: String,
-        updateKeyReference: String,
-        name: String
-    ): Identifier {
+    private fun generateRandomKeyId(): String {
+        return UUID.randomUUID().toString().replace("-", "")
```
beejones, line 88: Better to always use cryptographic random generators. I hope this is not used for key generation.

sdk/src/main/java/com/microsoft/did/sdk/crypto/CryptoOperations.kt
```diff
@@ -5,199 +5,95 @@
...
- * @param subtleCrypto primitives for operations.
- * @param keyStore specific keyStore that securely holds keys.
- */
-class CryptoOperations(
-    subtleCrypto: SubtleCrypto,
-    val keyStore: KeyStore,
-    private val ellipticCurvePairwiseKey: EllipticCurvePairwiseKey
-) {
-    val subtleCryptoFactory = SubtleCryptoFactory(subtleCrypto)
+    fun sign(payload: ByteArray, signingKey: PrivateKey, alg: SigningAlgorithm): ByteArray {
```
gproanomsft, line 29: consider operation specific terminology

you sign and verify digests.  at a glance payload makes me think this is a pre-image that gets hashed (and maybe that's what it is...so use preimage)

sdk/src/main/java/com/microsoft/did/sdk/crypto/CryptoOperations.kt
```diff
@@ -5,199 +5,95 @@
...
-    fun verify(payload: ByteArray, signature: ByteArray, signingKeyReference: String, algorithm: Algorithm? = null) {
-        SdkLog.d("Verifying with $signingKeyReference")
-        val publicKey = keyStore.getPublicKey(signingKeyReference)
-        val alg = algorithm ?: publicKey.alg ?: throw KeyException("No Algorithm specified for key $signingKeyReference")
-        val subtle = subtleCryptoFactory.getMessageSigner(alg.name, SubtleCryptoScope.PUBLIC)
-        val key = subtle.importKey(KeyFormat.Jwk, publicKey.getKey().toJWK(), alg, true, listOf(KeyUsage.Verify))
-        if (!subtle.verify(alg, key, signature, payload)) {
-            throw SignatureException("Signature invalid")
-        }
+    fun digest(payload: ByteArray, alg: DigestAlgorithm): ByteArray {
```
gproanomsft, line 51: payload = preimage, if what i'm assuming here this is hashing a preimage

sdk/src/main/java/com/microsoft/did/sdk/crypto/CryptoOperations.kt
```diff
@@ -5,199 +5,95 @@
...
+            if (alg.provider == null) MessageDigest.getInstance(alg.name) else MessageDigest.getInstance(alg.name, alg.provider)
+        return messageDigest.digest(payload)
     }
 
-    /**
-     * Encrypt payload with key stored in keyStore.
-     */
-    fun encrypt() {
-        TODO("Not implemented")
+    fun encrypt(payload: ByteArray, key: SecretKey, alg: CipherAlgorithm): ByteArray {
```
gproanomsft, line 57: same  here, consider plainText/cipherText for encrypt/decrypt

sdk/src/main/java/com/microsoft/did/sdk/crypto/CryptoOperations.kt
```diff
@@ -5,199 +5,95 @@
...
-        )
-        val key = crypto.importKey(
-            KeyFormat.Jwk, masterJwk, alg, false, listOf(
-                KeyUsage.Sign
-            )
-        )
-        return crypto.sign(alg, key, userDid.map { it.toByte() }.toByteArray())
+    fun generateSeed(): ByteArray {
+        val randomNumberGenerator = SecureRandom()
+        return randomNumberGenerator.generateSeed(16)
```
gproanomsft, line 97: 16 bytes is 128 bits, either the comment on line 93 is wrong, or this line is

sdk/src/main/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/JwsToken.kt
```diff
@@ -1,259 +1,72 @@
...
+import com.nimbusds.jose.jwk.JWK
+import com.nimbusds.jose.shaded.json.parser.ParseException
+import com.nimbusds.jose.util.Base64URL
+import java.security.PublicKey
 
-/**
- * Class for containing JWS token operations.
- * @class
- */
 class JwsToken private constructor(
```
gproanomsft, line 15: this class seems to be missing a alg property in the header, perhaps its set elsewhere?

sdk/src/main/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/JwsToken.kt
```diff
@@ -1,259 +1,72 @@
...
-        return if (all) {
-            results.reduce { result, valid ->
-                result && valid
-            }
-        } else {
-            results.reduce { result, valid ->
-                result || valid
+    fun verify(publicKeys: List<PublicKey> = emptyList()): Boolean {
+        for (key in publicKeys) {
+            val verifier = DefaultJWSVerifierFactory().createJWSVerifier(jwsObject.header, key)
```
gproanomsft, line 58: how efficient is it to create a public key every time you verify?  same with private key

sdk/src/main/java/com/microsoft/did/sdk/crypto/protocols/jose/jws/serialization/JwkSerializer.kt
gproanomsft: agree here, especially as i see 10 rsaJwk null compares

sdk/src/test/java/android/util/Base64.kt
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Android classes are not available in a test environment. This class serves as a mock of the android.util class.
+ *
+ * Tests will automatically use this Base64 encoder.
+ *
+ * CAUTION: Not all available methods and flags are implemented.
+ */
+object Base64 {
+    @JvmStatic
+    fun encodeToString(input: ByteArray?, flags: Int): String {
```
gproanomsft, line 18: flags is never used ?

sdk/src/test/java/com/microsoft/did/sdk/crypto/AlgorithmsTest.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun `test KeyGenAlgorithm Secp256k1`() {
+        val actualKeys = CryptoOperations.generateKeyPair(KeyGenAlgorithm.Secp256k1)
+        assertThat(actualKeys.private).isInstanceOf(ECPrivateKey::class)
+        assertThat(actualKeys.public).isInstanceOf(ECPublicKey::class)
+    }
+
+    @Test
+    fun `test MacAlgorithm HmacSha512`() {
+        val seed = ByteArray(16, { it.toByte() })
+        val secretKey = SecretKeySpec(seed, AES_KEY)
```
gproanomsft, line 36: line 35 and 36 are odd; it looks like you create a 16 byte seed that gets passed into some KDF to create another byte array of unspecified length.  which if it's not a 64 byte array, will just get a SHA-512 hash to create the HMAC key

why go through all this instead of defining a test HS512 key?

sdk/src/test/java/com/microsoft/did/sdk/identifier/SideTreeHelperTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val sideTreeHelper = SideTreeHelper()
+
+    private val testJwk = "{\"kty\":\"EC\",\"crv\":\"secp256k1\",\"kid\":\"signKey\",\"x\":\"dqEeSHC5KhsMSW_Zh8kBzQXB9HLgZqzBtmkAh-tAw4U\",\"y\":\"Yo_a4_sB2METsA9YRD6II_PjbHiWg4gwqQJiOxx4Suk\"}"
+    private val testJwkRearrangedWithSpaces = "{ \"kid\" : \"signKey\",\"crv\":\"secp256k1\",\"kty\":\"EC\",\"x\":\"dqEeSHC5KhsMSW_Zh8kBzQXB9HLgZqzBtmkAh-tAw4U\",\"y\":\"Yo_a4_sB2METsA9YRD6II_PjbHiWg4gwqQJiOxx4Suk\"}"
```
gproanomsft, line 15: in general i suggest a better solution than hard coding json structures, most of these look like they have classes that can be roundtripped to json, so for readability, consider using instantiating an instance and then using it to get json

gproanomsft (changes requested):

sdk/src/main/java/com/microsoft/did/sdk/identifier/IdentifierCreator.kt
AndiMiko: The random keyId is not cryptographically relevant. Indeed, we have set it deterministically before. An Sidetree observer might deduct information about the DID, like which client it is using and how many keys they own. I thought this might cause issues and instead made it random to prevent such attacks. Nonetheless from `randomUUID()` method comment:
```
     * The {@code UUID} is generated using a cryptographically strong pseudo
     * random number generator.
```

sdk/src/main/java/com/microsoft/did/sdk/identifier/SideTreeHelper.kt
AndiMiko: done

sdk/src/test/java/android/util/Base64.kt
AndiMiko: This class is a mock for tests only. We currently only ever use one flag anywhere.

I added a throw such that in case we would ever use other flags in tests it would not silently do the wrong thing and instead let the dev know.

sdk/src/test/java/com/microsoft/did/sdk/identifier/SideTreeHelperTest.kt
AndiMiko: are you suggesting to create some sort of JSON object here, and then serialize it? Part of the crux of these tests here is that a JSON string is passed and properly canonicalized. A JSON serializer would likely create a canonicalized string (or at least always in the same shape).

sdk/src/test/java/com/microsoft/did/sdk/crypto/AlgorithmsTest.kt
AndiMiko: What do you mean by defining a test HS512 key? Isn't defining a `SecretKeySpec` with a deterministic input just that?

Other options in Java include using a `KeyGenerator` to create such a key (which randomly generates a key and as such is not viable for tests) or `KeyFactory` which expects key material to be passed (a `SecretKeySpec`).

Not an expert on this so maybe I'm misunderstanding what you are suggesting.

sdk/src/androidTest/java/com/microsoft/did/sdk/crypto/keyStore/EncryptedKeyStoreInstrumentedTest.kt
beejones: This is a misted opportunity during a refactoring because we know we will need versions of keys. Also ION will need address this,

sdk/src/main/java/com/microsoft/did/sdk/crypto/CryptoOperations.kt
beejones: In the spec we use a 256 bit seed for pairwise key generation.

AndiMiko: This is the masterseed not directly the pairwise seed. I have changed the seed to 256 bit as per our conversation. Thanks for catching this.

nithyaganeshng (approved):

PR #149, opened 2021-03-02
handling 400 http code from service and including exception messages …

nithyaganeshng:
**Problem:**

1. When there is a 400 http status coming from service, it was being lost as it is not handled. 
2. When there is an unknown exception from underlying calls, a generic message is bubbled up but the exact reason for failure is lost.

**Solution:**
Handled 400 code to throw an exception.
When there is an unknown exception, its exception message is bubbled up along with the generic message from SDK.

**Validation:**

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:


sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
```diff
@@ -42,7 +42,7 @@ abstract class BaseNetworkOperation<S, T> {
             }
             return onFailure(response)
         } catch (exception: IOException) {
-            return Result.Failure(LocalNetworkException("Failed to send request.", exception))
+            return Result.Failure(LocalNetworkException("Failed to send request due to ${exception.message}", exception))
```
AndiMiko, line 45: Isn't the exception message printed also just below in the stack trace? Maybe as a convinience we can also add it here. Another option would be to just pass `exception` to `Result.Failure`, so we don't repack the exception into a `LocalNetworkException`. Not sure if there is a benefit of repacking it.

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Result.kt
```diff
@@ -42,7 +42,7 @@ suspend fun <T> runResultTry(block: suspend RunResultTryContext.() -> Result<T>)
         Result.Failure(ex)
     } catch (ex: Exception) {
         SdkLog.e("Unhandled Sdk Exception", ex)
-        Result.Failure(SdkException("Unhandled Exception", ex))
+        Result.Failure(SdkException("Unhandled Exception due to ${ex.message}", ex))
```
AndiMiko, line 45: same here. Should we just pass `ex`?

AndiMiko (approved):

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
nithyaganeshng: Cause is added to the exception but we are not actually using it anywhere. We always use just the exception message (repacked exception) and hence the actual cause was lost. As such we cannot add the exception directly to Failure as it is configured to accept only SdkException and that makes sense. Do we want to make the wallet use the cause then and not change anything here?

AndiMiko: The problem is that there can be (and often are) many causes in a loop. It's not straight forward to determine which of the causes is the "insightful" one. Your addition with "due to" defaults to the first cause, but that's just a heuristic in a sense. I would argue the last cause is often most helpful, but yeah, not always :D

nithyaganeshng: Yes that's true. But the message is something that is displayed to user during failure. Cause should be logged so that we can request them in case we couldn't debug from just the message. The failure could happen during interaction with any component or library so don't think we could use a well-guessed number to restrict the processed level of cause. Also not always clear on which one makes sense. But we should handle the errors coming from service to the best we could so that this will always be something unknown. In that case, outermost cause would be enough to understand the reason of failure for user. What do you think?

AndiMiko: I guess I was wondering whether repacking the exception into a `LocalNetworkException` is any different then, since you are basically pulling the message up from the exception below. Maybe instead just pass the original exception, done.

Not a big deal either way though.

nithyaganeshng: I am actually rethinking this. You are right repacking doesn't serve any purpose, it is sent to wallet. I am thinking of making the change on wallet side to do the same. Extract the top most message from cause in case of SdkException.

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Result.kt
nithyaganeshng: Removed these changes.

PR #152, opened 2021-03-11
Improve logging and telemetry event tracking

AndiMiko:
**Type of change:**
- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [X] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:


sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
```diff
@@ -2,12 +2,48 @@
...
+    const val TIME = "time"
+}
 
 inline fun <R> logTime(name: String, block: () -> R): R {
     val start = System.currentTimeMillis()
     val result = block()
     val elapsedTime = System.currentTimeMillis() - start
-    SdkLog.i("Perf: $name - ${elapsedTime}ms")
+    SdkLog.event(
+        "PerformanceMetrics", mapOf(
```
nithyaganeshng, line 20: could we add a prefix to all our events?

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
```diff
@@ -2,12 +2,48 @@
...
+    val cvRequest = result.raw().request().headers()[Constants.CORRELATION_VECTOR_HEADER] ?: "none"
+    val cvResponse = result.raw().headers()[Constants.CORRELATION_VECTOR_HEADER] ?: "none"
+    val requestId = result.raw().headers()[Constants.REQUEST_ID_HEADER]
+
+    SdkLog.event(
+        "NetworkMetrics", mapOf(
+            NAME to name,
+            TIME to "$elapsedTime",
+            "CV_request" to cvRequest,
+            "CV_response" to cvResponse,
```
nithyaganeshng, line 42: We don't actually use the CV in response, but I like the idea of having it if needed for debugging.

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
```diff
@@ -2,12 +2,48 @@
...
+}
+
+inline fun <S> logNetworkTime(name: String, block: () -> Response<S>): Response<S> {
+    val start = System.currentTimeMillis()
+    val result = block()
+    val elapsedTime = System.currentTimeMillis() - start
+
+    val cvRequest = result.raw().request().headers()[Constants.CORRELATION_VECTOR_HEADER] ?: "none"
+    val cvResponse = result.raw().headers()[Constants.CORRELATION_VECTOR_HEADER] ?: "none"
+    val requestId = result.raw().headers()[Constants.REQUEST_ID_HEADER]
```
nithyaganeshng, line 35: I believe we don't always have request id in the headers. For example, we don't have it in VC SDK but we do in services. I might be wrong. Don't we want to handle those in case it is not present?

sdk/src/main/java/com/microsoft/did/sdk/util/NetworkErrorParser.kt
```diff
@@ -0,0 +1,39 @@
...
+     *          or empty if not a valid json
+     */
+    fun extractInnerErrorsCodes(errorBody: String?): String? {
+        if (errorBody == null) return null
+        val errorCodes = ArrayList<String>()
+        try {
+            val json = Json.decodeFromString<JsonObject>(errorBody)
+            var error = (json["error"] as? JsonObject)
+            var code = (error?.get("code") as? JsonPrimitive)?.content
+            while (code != null) {
```
nithyaganeshng, line 29: Do we peel all the layers of inner errors? I thought we decided on 3 levels for now.

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
```diff
@@ -50,19 +50,25 @@ class RegistrarException(message: String, cause: Throwable? = null) : SdkExcepti
 
 open class LocalNetworkException(message: String, cause: Throwable? = null) : SdkException(message, cause, true)
 
-open class NetworkException(errorCode: String, message: String, retryable: Boolean, val requestId: String? = null, val correlationVector: String? = null) : SdkException(message, null, retryable, errorCode)
+open class NetworkException(message: String, retryable: Boolean) : SdkException(message, null, retryable) {
+    var requestId: String? = null
+    var correlationVector: String? = null
+    var errorCode: String? = null
+    var errorBody: String? = null
+    var innerErrors: String? = null
```
nithyaganeshng, line 58: So inner error codes don't replace the code we had already? So our order of preference in exposing them is inner error if present, else http code else exception name. Is that right? I haven't looked at wallet PR so just confirming.

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
```diff
@@ -56,65 +56,21 @@ abstract class BaseNetworkOperation<S, T> {
...
     // TODO("what do we want our base to look like")
     open fun onFailure(response: Response<S>): Result<Nothing> {
-        val requestId = response.headers()[REQUEST_ID_HEADER]
-        val correlationVector = response.headers()[CORRELATION_VECTOR_HEADER]
-        return when (response.code()) {
-            301, 302, 308 -> {
-                val errorMessage = StringBuilder("Redirecting is disabled: Http code: ${response.code()}\t")
-                if (response.errorBody() != null)
-                    errorMessage.append("ErrorBody: ${response.errorBody()?.string()}")
-                SdkLog.d(errorMessage.toString())
```
nithyaganeshng, line 66: We need the debug statement for redirect cases alone. Especially while fetching well-known config document, as it we don't bubble the failure in that case. This helps to figure what was wrong with the configuration without having to reproduce it.

sdk/src/test/java/com/microsoft/did/sdk/util/NetworkErrorParserTest.kt
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Microsoft Corporation. All rights reserved
+
+package com.microsoft.did.sdk.util
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class NetworkErrorParserTest {
```
nithyaganeshng, line 8: 👍

nithyaganeshng (changes requested):
It looks good to me. only have a few suggestions and questions.

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
AndiMiko: are you saying PICs is not responding with a CV in header?

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
AndiMiko: good catch

sdk/src/main/java/com/microsoft/did/sdk/util/NetworkErrorParser.kt
AndiMiko: It's more code to limit it to 3 and I don't see a reason why limiting it would be beneficial (it's not like a json can have infinite loops). I also don't think that service ever goes past depth 2 :D

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
AndiMiko: The reason why I didn't replace the message with the new innerErrors is because for the user we want to show the `errorBody` for debugging which also contains the messages etc. while for telemetry we want `innerErrors`. The wallet should make this decision conciously. The message can be anything but should be PII free so it's always safe to log

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
AndiMiko: done

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
AndiMiko: why will the error not bubble up for redirects only? If I understand what you are saying it should apply to every possible failure, doesn't it?

nithyaganeshng: I meant it doesn't bubble up in fetching well-known config doc scenario. It is by design because we treat it like a warning and not failure. So we log it and show a warning that it wasn't successful. Only from the log statement we know what is the failure.

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
nithyaganeshng: PICs sends it but we don't use it.

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
nithyaganeshng: I was referring to error code but its okay if we don't replace it as we  would need both this and error code to be diaplayed.

sdk/src/main/java/com/microsoft/did/sdk/util/NetworkErrorParser.kt
nithyaganeshng: Our service doesn't do many layers but we decided to limit it to 3 because we don't know about the services SDK will be talking to. And if it is more, it becomes really long to display it in the wallet to the user.

sdk/src/main/java/com/microsoft/did/sdk/util/MeasureTime.kt
AndiMiko: Isn't the CV supposed to count up the CV that it received - now that I think about it? We only count up internally, right?

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/BaseNetworkOperation.kt
AndiMiko: Yes, but wouldn't the same argument apply to a 404 on a well-known fetch?

nithyaganeshng (approved):

PR #189, opened 2021-10-28
Niganesh/multiple to single vp

nithyaganeshng:
- Move from single VP for a single VC to single VP for multiple VCs.
- Change the structure of Presentation Submission in response according to latest OIDC4VP spec.

**Validation:**

**Type of change:**
- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:


sdk/src/main/java/com/microsoft/did/sdk/credential/service/protectors/PresentationResponseFormatter.kt
```diff
@@ -30,77 +30,67 @@ class PresentationResponseFormatter @Inject constructor(
         presentationResponse: PresentationResponse,
         responder: Identifier,
         expiryInSeconds: Int = Constants.DEFAULT_EXPIRATION_IN_SECONDS
-    ): String {
+    ): Pair<String, String> {
```
symorton, line 33: Instead of using Pair<>, should we create a new object with more detail about what the first and second string actually is? So maybe a SignedResponseContents object with the oidcResponseClaims and attestions in it? Or maybe we should have a VerifiablePresentationFormatter separate from the PresentationResponseFormatter in general and inject both formatters in the PresentationService? what do you think?

sdk/src/main/java/com/microsoft/did/sdk/util/Constants.kt
```diff
@@ -20,9 +20,9 @@ object Constants {
     const val ALGORITHM_SUPPORTED_IN_VP = "ES256K"
 
     //Presentation Exchange Constants
-    const val CREDENTIAL_PATH_IN_RESPONSE = "$.attestations.presentations"
-    const val CREDENTIAL_PRESENTATION_FORMAT = "JWT"
-    const val CREDENTIAL_PRESENTATION_ENCODING = "base64Url"
+    const val CREDENTIAL_PATH_IN_RESPONSE = "$.verifiableCredential["
```
symorton, line 23: is there supposed to be a [ there? just double checking haha

PR #201, opened 2022-05-09
Have exception handling in the service class to throw specific exceptions depending on the scenario.

nithyaganeshng:
**Problem:**
Generic exceptions are exposed to the user which doesn't provide specific error or action item for them.

**Solution:**
Throw specific exceptions in the service class which can be used to distinguish the failure scenarios.

**Validation:**

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:

### DIFF
--- a/sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
@@ -19,10 +19,6 @@ import com.microsoft.did.sdk.datasource.network.credentialOperations.SendVerifia
 import com.microsoft.did.sdk.identifier.models.Identifier
 import com.microsoft.did.sdk.internal.ImageLoader
 import com.microsoft.did.sdk.util.Constants
-import com.microsoft.did.sdk.util.controlflow.ExpiredTokenException
-import com.microsoft.did.sdk.util.controlflow.ForbiddenException
-import com.microsoft.did.sdk.util.controlflow.InvalidPinException
-import com.microsoft.did.sdk.util.controlflow.NotFoundException
 import com.microsoft.did.sdk.util.controlflow.Result
 import com.microsoft.did.sdk.util.controlflow.runResultTry
 import com.microsoft.did.sdk.util.logTime
@@ -110,31 +106,12 @@ class IssuanceService @Inject constructor(
             responder = responder,
             expiryInSeconds = expiryInSeconds
         )
-        val result = SendVerifiableCredentialIssuanceRequestNetworkOperation(
+        return SendVerifiableCredentialIssuanceRequestNetworkOperation(
             response.audience,
             formattedResponse,
             apiProvider,
             jwtValidator,
             serializer
         ).fire()
-        when (result) {
-            is Result.Failure -> {
-                when (val sdkException = result.payload) {
-                    is ForbiddenException -> {
-                        val exception = InvalidPinException("Incorrect verification code", false)
-                        exception.apply {
-                            correlationVector = sdkException.correlationVector
-                            errorBody = sdkException.errorBody
-                            errorCode = sdkException.errorCode
-                            innerErrorCodes = sdkException.innerErrorCodes
-                            errorMessage = sdkException.errorMessage
-                        }
-                        return Result.Failure(exception)
-                    }
-                }
-            }
-            is Result.Success -> return result
-        }
-        return result
     }
 }
--- a/sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
+++ b/sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
@@ -19,6 +19,10 @@ import com.microsoft.did.sdk.datasource.network.credentialOperations.SendVerifia
 import com.microsoft.did.sdk.identifier.models.Identifier
 import com.microsoft.did.sdk.internal.ImageLoader
 import com.microsoft.did.sdk.util.Constants
+import com.microsoft.did.sdk.util.controlflow.ExpiredTokenException
+import com.microsoft.did.sdk.util.controlflow.ForbiddenException
+import com.microsoft.did.sdk.util.controlflow.InvalidPinException
+import com.microsoft.did.sdk.util.controlflow.NotFoundException
 import com.microsoft.did.sdk.util.controlflow.Result
 import com.microsoft.did.sdk.util.controlflow.runResultTry
 import com.microsoft.did.sdk.util.logTime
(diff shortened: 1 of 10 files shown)


sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
```diff
@@ -106,12 +110,31 @@ class IssuanceService @Inject constructor(
...
+                            innerErrorCodes = sdkException.innerErrorCodes
+                            errorMessage = sdkException.errorMessage
+                        }
+                        return Result.Failure(exception)
+                    }
+                }
+            }
+            is Result.Success -> return result
+        }
+        return result
```
AndiMiko, line 138: Instead of processing the result here after the fact (in theory you would need to add this code anywhere where this call is done) move this to on override method `onFailure` of `SendVerifiableCredentialIssuanceRequestNetworkOperation`. As the first line you can (should) call it's super and then process it more individually.

sdk/src/main/java/com/microsoft/did/sdk/PresentationService.kt
```diff
@@ -85,8 +88,28 @@ class PresentationService @Inject constructor(
         return Result.Success(serializer.decodeFromString(PresentationRequestContent.serializer(), jwsToken.content()))
     }
 
-    private suspend fun fetchRequest(url: String) =
-        FetchPresentationRequestNetworkOperation(url, apiProvider, jwtValidator, serializer).fire()
+    private suspend fun fetchRequest(url: String): Result<PresentationRequestContent> {
+        val result = FetchPresentationRequestNetworkOperation(url, apiProvider, jwtValidator, serializer).fire()
```
AndiMiko, line 92: same as other comment

sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
```diff
@@ -106,12 +110,31 @@ class IssuanceService @Inject constructor(
...
             formattedResponse,
             apiProvider,
             jwtValidator,
             serializer
         ).fire()
+        when (result) {
+            is Result.Failure -> {
+                when (val sdkException = result.payload) {
+                    is ForbiddenException -> {
+                        val exception = InvalidPinException("Incorrect verification code", false)
```
AndiMiko, line 124: this implies that every HTTP 403 response is an invalidPin? I don't think thats true (at best it is for now), I guess that logic needs to read the error codes instead?

sdk/src/main/java/com/microsoft/did/sdk/PresentationService.kt
```diff
@@ -85,8 +88,28 @@ class PresentationService @Inject constructor(
...
+        when (result) {
+            is Result.Failure -> {
+                when (val sdkException = result.payload) {
+                    is NotFoundException -> {
+                        val exception = ExpiredTokenException("The request has expired.", false)
```
AndiMiko, line 97: same as other comment

sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
AndiMiko: I think you probably want to base it on a check with `innerErrors` (that was the purpose of parsing it, but we haven't utilized it yet ..)

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
```diff
@@ -84,7 +82,8 @@ open class NetworkException(message: String, retryable: Boolean) : SdkException(
     var correlationVector: String? = null
     var errorCode: String? = null
     var errorBody: String? = null
-    var innerErrors: String? = null
+    var innerErrorCodes: String? = null
+    var errorMessage: String? = null
```
AndiMiko, line 86: all exceptions already have a (error)`message` field. Maybe just use that? So far we have only stored the entire responseBody there, but it is also stored in `errorBody` already, so feel free to use message for what you are doing (unless I'm missing something)

AndiMiko: but maybe make sure in the error drawer we would also like to display the entire responseBody as well? Just in case the parsing is failing. So show the "extracted" messages to the user and the responseBody to developers

AndiMiko (changes requested):

sdk/src/main/java/com/microsoft/did/sdk/IssuanceService.kt
nithyaganeshng: That is the plan. Already spoke to Sydney about it yesterday and she is going to start a thread with final error codes that could be used.

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
AndiMiko: I see what you are saying, but I think you missunderstood me. `responseBody` is already stored in `errorBody`. So you can use the default message and still have both information available at all times.

sdk/src/main/java/com/microsoft/did/sdk/PresentationService.kt
nithyaganeshng: We are doing it this way for now for this scenario.

sdk/src/main/java/com/microsoft/did/sdk/util/controlflow/Exceptions.kt
nithyaganeshng: got it. done.

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/credentialOperations/FetchPresentationRequestNetworkOperation.kt
```diff
@@ -31,6 +33,23 @@ class FetchPresentationRequestNetworkOperation(
         return verifyAndUnwrapPresentationRequest(jwsTokenString)
     }
 
+    override fun onFailure(response: Response<String>): Result<Nothing> {
+        val result = super.onFailure(response)
+        when (val exception = (result as Result.Failure).payload) {
+            is NotFoundException -> {
+                val expiredTokenException = ExpiredTokenException(exception.message ?: "", false)
```
AndiMiko, line 40: this is still missing the check for innerError?

nithyaganeshng: yes, this is how it is going to be checked for now until something changes on service side.

AndiMiko: This will show a wrong error message when the token is not expired but a 403 is returned. I think this is much more missleading than the other way around where we show a generic error instead for both cases

nithyaganeshng: We check for a 404 in this case and it happens when the request has expired or has already been scanned. For either of those scenarios, request expired page seems to be okay.

AndiMiko: yeah sorry I meant 404. Still could be other cases as well for a 404. We can leave this for now, but please let the service team know that we need specific errorcodes for this in the near future.

AndiMiko (approved):

sdk/src/main/java/com/microsoft/did/sdk/datasource/network/credentialOperations/FetchPresentationRequestNetworkOperation.kt
nithyaganeshng: will do.

PR #239, opened 2023-07-11
Skip fetching logo if it fails and proceed with the flow.

nithyaganeshng:
If fetching the logo fails during an issuance/presentation flow, skip it and proceed with the flow.

**Validation:**

**Type of change:**
- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

**Risk**:

**Work Item links**:

**Documentation Links**:


symorton (approved):
Approved, but why not just catch the error in the Auth App, and show missing label instead of returning nil here? So that we don't have to make changes in the SDK?

nithyaganeshng:
> Approved, but why not just catch the error in the Auth App, and show missing label instead of returning nil here? So that we don't have to make changes in the SDK?

This is called in SDK and populated in the request while getting the request. Auth App will not know which network call failed.

